PR #1, opened 2019-11-05
SREP-2251: Add webhook validation framework

lisa:
Provides a method for adding webhooks and injecting generated caBundles for use by said webhook. Refer to the README.md for additional details on how to add new hooks.

The included hook will address the group validation request for SREP-2251.

Things which need to change:

* Where the image lives (it shouldn't live in my personal namespace)
* Makefile & friends once the CI/CD pieces are connected.


lisa:
Note: https://github.com/openshift/service-ca-operator/pull/79 will change how this functions by removing the need for the "injector" script.

templates/05-group-validation-webhook.service.yaml.tmpl
```diff
@@ -0,0 +1,16 @@
+apiVersion: v1
```
rogbas: We're having a single service for all the webhooks, right? If so, it seems to me that we shouldn't have `group-validation` in the file name

templates/20-group-validation-webhook.deployment.yaml.tmpl
```diff
@@ -0,0 +1,59 @@
+apiVersion: extensions/v1beta1
```
rogbas: Same comment from the service object, since it will be a single deployment for all the webhooks

deploy/02-webhook.permissions.yaml
```diff
@@ -0,0 +1,39 @@
...
+metadata:
+  name: webhook-validation-cr
+rules:
+- apiGroups:
+  - admissionregistration.k8s.io
+  resources:
+    - validatingwebhookconfigurations
+  verbs:
+    - list
+    - patch
```
rogbas: Is there any use case that requires this `patch` permission on VWC

templates/10-group-validation.ValidatingWebhookConfiguration.yaml.tmpl
```diff
@@ -0,0 +1,52 @@
+apiVersion: admissionregistration.k8s.io/v1beta1
+kind: ValidatingWebhookConfiguration
+metadata: 
+  name: sre-validate-webhook
```
rogbas: Shouldn't `metadata.name` refer to group-validation?

templates/05-group-validation-webhook.service.yaml.tmpl
lisa: Good call.

templates/20-group-validation-webhook.deployment.yaml.tmpl
lisa: Agreed 👍

deploy/02-webhook.permissions.yaml
lisa: The initContainer needs `patch` to inject the caBundle.

templates/10-group-validation.ValidatingWebhookConfiguration.yaml.tmpl
lisa: Yep. It should also have only one webhook 😊 I'll fix both.

deploy/02-webhook.permissions.yaml
lisa: What we could do is limit the resourceNames to some list of specified (or computed, from yaml) VWC. I can spend some effort on that if you think it's worthwhile, otherwise, all this service account will be able to edit all VWC.

lisa:
We should also take into consideration what it means to have a different priority for the Deployment so it doesn't get evicted.

deploy/02-webhook.permissions.yaml
rogbas: I think it's ok for now, just confirming since I couldn't think of a use case for patch

PR #3, opened 2019-11-08
Needing a build and pr check script added to the webhook repo

robotmaxtron:
Adding a pr check script and a build deploy script

### DIFF
--- a/build/build_deploy.sh
+++ b/build/build_deploy.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+
+# AppSRE team CD
+
+set -exv
+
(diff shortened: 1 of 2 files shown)


build/build_deploy.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+# AppSRE team CD
+
+set -exv
+
+CURRENT_DIR=$(dirname "$0")
+
+BASE_IMG="tbd"
```
rogbas: IMO you could name the base image after the repo: `managed-cluster-validating-webhooks`

PR #4, opened 2019-11-08
Ensure the deploy directory exists

lisa:


lisa:
/hold

i should change this PR to also delete the contents of `deploy`, and ensure it's added to `.gitignore` as well.

jharrington22:
@lisa why are we removing the deploy folder here?

lisa:
the contents are provided by `make render`, or #7 perhaps.

rogbas:
@jharrington22 the selectorSyncSet is being generated from the files in `templates/`, created/populated in #7

Makefile
```diff
@@ -21,6 +21,7 @@ build-base: build/Dockerfile
 # TODO: Change the render to allow for the permissions to have a list of all the webhook names
 # TODO: Pull that list of names from the yaml files?
 render: $(TEMPLATEFILES) build/Dockerfile
+	mkdir deploy || true ;\
```
2uasimojo, line 24: This will spuriously pass if `mkdir deploy` fails for reasons other than the directory already existing (e.g. `EPERM`, or there's a regular file called `deploy` in the way). Better to use `mkdir -p deploy` and no `|| true`, since you presumably want to abort if one of those other things is wrong. (Um, I'm only guessing make targets implicitly run with `-e`?)

lisa: @2uasimojo That's a really good point. I think this PR is pretty stale, so it can be closed. I'll certainly keep an eye on this sort of thing.

PR #6, opened 2019-11-14
SREP-2239: Add Subscription validation

lisa:
This changeset adds a validation webhook using the same pattern as in #1
which will constrain dedicated-admins to making `Subscription` objects
whose `.spec.sourceNamespace` is defined in a comma-separated list from
the `SUBSCRIPTION_VALIDATION_N    AMESPACES` environment variable (By
default it is only the `openshift-operators` namespace whic is a
permitted `.spec.sourceNamespace` for dedicated-admins).

This changeset also makes some other structural changes:

* Move the environment variables relating to the webhooks out of the
gunicorn.py and into the `Deployment` spec (so that it makes more sense)
* Fundamental changes to the structure of the validation library so that
it is now only looking for the basic structure of the AdmissionReview
(does it have a `kind`? Is it `AdmissionReview`). This ripples on into
the validator hooks as they now ought to use try/except blocks to
account for missing elements in the incoming document.
* Changes the group validator to use the above try/except, and to use sensible
defaults for the two inputs it needs (group prefix, and admin group).
Both can still be overridden by environment variables.

To test the Subscription check, these two documents will put the code
through its paces:

Should fail:

```yaml
apiVersion: operators.coreos.com/v1alpha1
kind: Subscription
metadata:
  name: should-fail
  namespace: myns
spec:
  channel: stable
  name: should-fail
  source: should-fail
  sourceNamespace: default
```

Should pass:

```yaml
apiVersion: operators.coreos.com/v1alpha1
kind: Subscription
metadata:
  name: should-pass
  namespace: myns
spec:
  channel: stable
  name: should-pass
  source: should-pass
  sourceNamespace: openshift-operators
```

Signed-off-by: Lisa Seelye <lseelye@redhat.com>

### DIFF
--- a/src/webhook/group_validation.py
+++ b/src/webhook/group_validation.py
@@ -1,48 +1,36 @@
 from flask import request, Blueprint
+import sys, traceback
 import json
 import os
 
 from webhook.request_helper import validate, responses
 
-bp = Blueprint("webhook", __name__)
+bp = Blueprint("group-webhook", __name__)
 
-group_prefix = os.getenv("GROUP_VALIDATION_PREFIX")
-admin_group = os.getenv("GROUP_VALIDATION_ADMIN_GROUP")
+group_prefix = os.getenv("GROUP_VALIDATION_PREFIX", "osd-sre-")
+admin_group = os.getenv("GROUP_VALIDATION_ADMIN_GROUP", "osd-sre-admins")
 
-configok = True
+@bp.route('/group-validation', methods=('GET','POST'))
+def handle_request():
+  debug = os.getenv("DEBUG_GROUP_VALIDATION", "False")
+  debug = (debug == "True")
 
-if group_prefix is None:
-  print("Please specify the group prefix with GROUP_VALIDATION_PREFIX environment variable. Refusing to expose /group-validation webhook.")
-  configok = False
+  if debug:
+    print("REQUEST BODY => {}".format(request.json))
 
-if admin_group is None:
-  print("Please specify the admin group with the GROUP_VALIDATION_ADMIN_GROUP environment variable. Refusing to expose /group-validation webhook.")
-  configok = False
+  valid = True
+  try:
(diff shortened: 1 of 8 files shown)


src/webhook/subscription_validation.py
```diff
@@ -0,0 +1,45 @@
...
+
+from webhook.request_helper import validate, responses
+
+bp = Blueprint("subscription-webhook", __name__)
+
+valid_source_namespaces = os.getenv("SUBSCRIPTION_VALIDATION_NAMESPACES", "openshift-operators")
+
+valid_source_namespaces = valid_source_namespaces.split(",")
+
+@bp.route('/subscription-validation', methods=('GET','POST'))
```
rogbas: Any reason for supporting the GET method as well? Considering the admission controller does a POST

lisa: Not particularly. (The group one accepts `GET`, too) Happy to 🔪 them.

src/webhook/group_validation.py
```diff
@@ -1,48 +1,36 @@
...
-bp = Blueprint("webhook", __name__)
+bp = Blueprint("group-webhook", __name__)
 
-group_prefix = os.getenv("GROUP_VALIDATION_PREFIX")
-admin_group = os.getenv("GROUP_VALIDATION_ADMIN_GROUP")
+group_prefix = os.getenv("GROUP_VALIDATION_PREFIX", "osd-sre-")
+admin_group = os.getenv("GROUP_VALIDATION_ADMIN_GROUP", "osd-sre-admins")
 
-configok = True
+@bp.route('/group-validation', methods=('GET','POST'))
```
lisa:
```suggestion
@bp.route('/group-validation', methods=('POST'))
```

src/webhook/subscription_validation.py
lisa:
```suggestion
@bp.route('/subscription-validation', methods=('POST'))
```

PR #7, opened 2019-11-15
Initial effort at selectorsyncset template

robotmaxtron:
As part of the CICD process for SREP-2105


build/build_deploy.sh
```diff
@@ -8,13 +8,17 @@ CURRENT_DIR=$(dirname "$0")
...
+IMG="${BASE_IMG}:${VERSION_MAJOR}.${VERSION_MINOR}-${GIT_HASH}"
+VERSION_MAJOR=0
+VERSION_MINOR=1
 
 GIT_HASH=$(git rev-parse --short=7 HEAD)
 
 # build the image
 BUILD_CMD="docker build" IMG="$IMG" make docker-build
 
+GEN_SYNCSET=$(generate_syncset.py -t "${SELECTOR_SYNC_SET_TEMPLATE_DIR}" -y "${YAML_DIRECTORY}" -d "${SELECTOR_SYNC_SET_DESTINATION}" -r "${REPO_NAME}")
```
lisa: Is this line meant to be here? I'm not seeing where these variables are set.

robotmaxtron: Variable definition added, thanks for the initial review/comment :+1:

build/build_deploy.sh
```diff
@@ -8,13 +8,20 @@ CURRENT_DIR=$(dirname "$0")
 
 BASE_IMG="managed-cluster-validating-webhooks"
 QUAY_IMAGE="quay.io/app-sre/${BASE_IMG}"
-IMG="${BASE_IMG}:latest"
-
+IMG="${BASE_IMG}:${VERSION_MAJOR}.${VERSION_MINOR}-${GIT_HASH}"
```
rogbas: Could we maybe have the same versioning scheme from the other osd operators?
MAJOR.MINOR.<commit count>-GIT_SHA

build/build_deploy.sh
rogbas: Would it make sense to check the return of the generate_syncset.py call, and explicitly error out the entire process in case it fails

build/build_deploy.sh
```diff
@@ -8,13 +8,20 @@ CURRENT_DIR=$(dirname "$0")
...
+VERSION_MAJOR=0
+VERSION_MINOR=1
+SELECTOR_SYNC_SET_TEMPLATE_DIR=templates
+YAML_DIRECTORY=deploy
+SELECTOR_SYNC_SET_DESTINATION=templates/00-osd-managed-cluster-validating-webhooks.selectorsyncset.yaml.tmpl
+REPO_NAME=managed-cluster-validating-webhooks
 GIT_HASH=$(git rev-parse --short=7 HEAD)
 
 # build the image
 BUILD_CMD="docker build" IMG="$IMG" make docker-build
```
rogbas: I'm not getting this make docker-build at the end of the docker build command.

Is it supposed to be here? If that's the case we don't have a `docker-build` handle in this repo's Makefile

build/build_deploy.sh
rogbas: I tried running the build_deploy.sh script manually and got a `build/build_deploy.sh: line 23: generate_syncset.py: command not found`
Should call the generate_syncset.py with the full path? including `build/`

build/generate_syncset.py
```diff
@@ -0,0 +1,117 @@
...
+    #Argument parser
+    parser = argparse.ArgumentParser(description="selectorsyncset generation tool", usage='%(prog)s [options]')
+    parser.add_argument("--template-dir", "-t", required=True, help="Path to template directory [required]")
+    parser.add_argument("--yaml-directory", "-y", required=True, help="Path to folder containing yaml files [required]")
+    parser.add_argument("--destination", "-d", required=True, help="Destination for selectorsynceset file [required]")
+    parser.add_argument("--repo-name", "-r", required=True, help="Name of the repository [required]")
+    arguments = parser.parse_args()
+
+    # Get the template data
+    template_data = get_yaml(os.path.join(arguments.template_dir, "template.yaml"))
```
rogbas: Are we supposed to have a `template.yaml` file in the `templates/` folder? 
I'm not finding it now, nvm my comment if it's coming in a following PR.

build/generate_syncset.py
```diff
@@ -0,0 +1,117 @@
...
+    selectorsyncset_data = get_yaml(os.path.join(arguments.template_dir, "selectorsyncset.yaml"))
```
rogbas: Should this file name be `00-osd-managed-cluster-validating-webhooks.selectorsyncset.yaml.tmpl` instead of `selectorsyncset.yaml`?

build/build_deploy.sh
```diff
@@ -8,19 +8,26 @@ CURRENT_DIR=$(dirname "$0")
 
 BASE_IMG="managed-cluster-validating-webhooks"
 QUAY_IMAGE="quay.io/app-sre/${BASE_IMG}"
-IMG="${BASE_IMG}:latest"
-
+VERSION_MAJOR=0
+VERSION_MINOR=1
+SELECTOR_SYNC_SET_TEMPLATE_DIR=../templates
+BUILD_DIRECTORY=../build
```
rogbas:
```suggestion
BUILD_DIRECTORY=build
```

build/build_deploy.sh
```diff
@@ -8,19 +8,26 @@ CURRENT_DIR=$(dirname "$0")
...
+REPO_NAME=managed-cluster-validating-webhooks
 GIT_HASH=$(git rev-parse --short=7 HEAD)
+IMAGETAG="${VERSION_MAJOR}.${VERSION_MINOR}-${GIT_HASH}"
+IMG="$QUAY_IMAGE":"$IMAGETAG"
+
+# build the image and the selectorsyncset
 
-# build the image
-BUILD_CMD="docker build" IMG="$IMG" make docker-build
+python generate_syncset.py -t ${SELECTOR_SYNC_SET_TEMPLATE_DIR} -b ${BUILD_DIRECTORY} -d ${SELECTOR_SYNC_SET_DESTINATION} -r ${REPO_NAME}
```
rogbas:
```suggestion
python build/generate_syncset.py -t ${SELECTOR_SYNC_SET_TEMPLATE_DIR} -b ${BUILD_DIRECTORY} -d ${SELECTOR_SYNC_SET_DESTINATION} -r ${REPO_NAME}
```

build/build_deploy.sh
```diff
@@ -8,19 +8,26 @@ CURRENT_DIR=$(dirname "$0")
...
 
+docker build -t $IMG -f Dockerfile ..
```
rogbas:
```suggestion
IMG="$QUAY_IMAGE" IMAGETAG="$IMAGETAG" make build-base
```

PR #9, opened 2019-11-22
SREP-2442: Allow kube:admin to create any group

lisa:
We may need to refactor this later to encapsulate more permissions, but
for now, kube:admin functions as "`root`", and can do anything.

Signed-off-by: Lisa Seelye <lseelye@redhat.com>

### DIFF
--- a/src/webhook/group_validation.py
+++ b/src/webhook/group_validation.py
@@ -31,6 +31,11 @@ def handle_request():
     body_dict = request.json['request']
     group_name = body_dict['object']['metadata']['name']
     userinfo = body_dict['userInfo']
+    if userinfo['username'] == "kube:admin":
+      # kubeadmin can do anything
+      if debug:
+        print("Performing action: {} in {} group for kube:admin".format(body_dict['operation'],group_name))
+      return responses.response_allow(req=body_dict)
     if group_name.startswith(group_prefix):
       if debug:
         print("Performing action: {} in {} group".format(body_dict['operation'],group_name))
--- a/src/webhook/group_validation.py
+++ b/src/webhook/group_validation.py
@@ -8,7 +8,9 @@ from webhook.request_helper import validate, responses
 bp = Blueprint("group-webhook", __name__)
 
 group_prefix = os.getenv("GROUP_VALIDATION_PREFIX", "osd-sre-")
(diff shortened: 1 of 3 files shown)


src/webhook/group_validation.py
```diff
@@ -31,6 +31,11 @@ def handle_request():
     body_dict = request.json['request']
     group_name = body_dict['object']['metadata']['name']
     userinfo = body_dict['userInfo']
+    if userinfo['username'] == "kube:admin":
```
rogbas: I'm not sure if the user will come as `kube:admin` or just `admin` in the AdmissionReview object. 
Were you able to confirm?

rogbas: The other group providing permissions higher than `osd-sre-admins` is `osd-sre-cluster-admins`, should we also address it here?

lisa: I can confirm that it's `kube:admin`. Here is a sample request body (forgive the Python format, but it's close enough to JSON):

```python
{
  'kind': 'AdmissionReview', 
  'apiVersion': 'admission.k8s.io/v1beta1', 
  'request':
  {
    'uid': 'd0f53355-0d1b-11ea-8394-0a580a820038', 
    'kind': 
    {
      'group': 'user.openshift.io', 
      'version': 'v1',
      'kind': 'Group'
    },
    'resource': 
    {
      'group': 'user.openshift.io',
      'version': 'v1',
      'resource': 
      'groups'
    }, 
    'operation': 
    'CREATE', 
    'userInfo': 
    {
      'username': 
      'kube:admin', 
      'groups': 
      [
        'system:cluster-admins', 
        'system:authenticated'
      ], 
      'extra': 
      {
        'scopes.authorization.openshift.io': 
        [
          'user:full'
        ]
      }
    }, 
    'object': 
    {
      'kind': 
      'Group', 
      'apiVersion': 
      'user.openshift.io/v1', 
      'metadata': 
      {
        'name': 'osd-sre-foo123', 
        'uid': 'd0f52f1d-0d1b-11ea-8394-0a580a820038', 
        'creationTimestamp': '2019-11-22T11:32:50Z'
      }, 
      'users': None
    }, 
    'oldObject': None, 
    'dryRun': False
  }
}
```

PR #16, opened 2019-12-13
More updates to correctly generate selectorsyncset.yaml

robotmaxtron:
As part of SREP-2105

### DIFF
--- a/templates/20-validation-webhook.deployment.yaml.tmpl
+++ b/templates/20-validation-webhook.deployment.yaml.tmpl
@@ -15,7 +15,7 @@ spec:
     spec:
       serviceAccountName: "#SANAME#"
       initContainers:
-      - image: "#IMG#:#IMAGETAG#"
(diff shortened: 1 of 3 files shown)


templates/20-validation-webhook.deployment.yaml.tmpl
```diff
@@ -15,15 +15,15 @@ spec:
     spec:
       serviceAccountName: "#SANAME#"
       initContainers:
-      - image: "#IMG#:#IMAGETAG#"
```
lisa: Does this work? I'm guessing the SelectorSyncSet template will do the substitution?

robotmaxtron: The easiest answer seemed to be updating the template to ${IMAGE_TAG}, though I think it could be cleaned up since we have several variables like IMAGETAG and  IMAGE_TAG that could be done better. The selectorsyncset.yaml file that's in the PR is the resulting output of the changes, including this one.

PR #18, opened 2019-12-19
OSD-2573 Disable group WH in stage during

rogbas:
Group webhook is blocking cluster provisioning in stage.

Disabling group WH during shutdown in stage, after change freeze it will be revisited.


robotmaxtron:
+1

PR #22, opened 2020-01-20
Expose metrics for validation webhooks

sam-nguyen7:
Added following metrics:

1. webhook_subscription_validation - The total number of subscription validation requests
2. webhook_denied_subscription - The total number of subscription validation requests denied
3. webhook_group_validation - The total number of group validation requests
4. webhook_denied_group - The total number of group validation requests denied

### DIFF
--- a/src/webhook/group_validation.py
+++ b/src/webhook/group_validation.py
@@ -2,10 +2,14 @@ from flask import request, Blueprint
 import sys, traceback
 import json
 import os
-
+from prometheus_client import Counter
 from webhook.request_helper import validate, responses
 
 bp = Blueprint("group-webhook", __name__)
+
+# define what we track, declare Counter, how many times this route is accessed
+TOTAL_GROUP = Counter('webhook_group_validation_total', 'The total number of group validation requests')
+DENIED_GROUP= Counter('webhook_group_validation_denied', 'The total number of group validation requests denied')
 
 group_prefix = os.getenv("GROUP_VALIDATION_PREFIX", "osd-sre-")
 admin_group = os.getenv("GROUP_VALIDATION_ADMIN_GROUP", "osd-sre-admins,osd-sre-cluster-admins")
@@ -14,6 +18,8 @@ admin_groups = admin_group.split(",")
 
 @bp.route('/group-validation', methods=['POST'])
 def handle_request():
+   # inc total group counter
+  TOTAL_GROUP.inc()
   debug = os.getenv("DEBUG_GROUP_VALIDATION", "False")
   debug = (debug == "True")
 
@@ -27,6 +33,8 @@ def handle_request():
     valid = False
 
   if not valid:
+    # inc denied group counter
+    DENIED_GROUP.inc()
     return responses.response_invalid()
 
   try:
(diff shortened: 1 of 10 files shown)


src/webhook/group_validation.py
```diff
@@ -2,18 +2,26 @@
 import sys, traceback
 import json
 import os
+import prometheus_client
+from prometheus_client import Counter
```
fahlmant: Do you need this since you import the whole package above?

templates/30-validation-webhook.servicemonitor.yaml.tmpl
```diff
@@ -0,0 +1,17 @@
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  labels:
+    name: "#SVCNAME#"
+  name: "#SVCNAME#"
+  namespace: "#NAMESPACE#"
+spec:
+  endpoints:
```
fahlmant: This servicemonitor is for an /https endpoint, but according to `metrics.py` the endpoint is `/metrics`

src/webhook/subscription_validation.py
```diff
@@ -1,18 +1,26 @@
...
+from prometheus_client import Counter
+from prometheus_client.core import CollectorRegistry
 
 from webhook.request_helper import validate, responses
 
 bp = Blueprint("subscription-webhook", __name__)
 
-valid_source_namespaces = os.getenv("SUBSCRIPTION_VALIDATION_NAMESPACES", "openshift-marketplace")
+# define what we track, declare Counter, how many times this route is accessed
+TOTAL_SUBSCRIPTION = Counter('webhook_subscription_validation', 'The total number of subscription validation requests')
```
rogbas: Same naming standard from the other webhook applies here.

templates/30-validation-webhook.servicemonitor.yaml.tmpl
```diff
@@ -0,0 +1,17 @@
...
+  - port: "https"
+    scheme: "https"
+    tlsConfig: 
+      insecureSkipVerify: true
```
rogbas: We'd need to replace the insecure bit with the actual TLS config before we go into production with that.

sam-nguyen7: @rogbas when you say production does that mean both environments (stage and prod) or just prod?

src/webhook/group_validation.py
sam-nguyen7: I did the inverse and it worked. deleting `import prometheus_client` and retaining `from prometheus_client import Counter`

src/webhook/group_validation.py
```diff
@@ -2,18 +2,26 @@
...
 
 from webhook.request_helper import validate, responses
 
 bp = Blueprint("group-webhook", __name__)
 
+# define what we track, declare Counter, how many times this route is accessed
+TOTAL_GROUP = Counter('webhook_group_validation', 'The total number of group validation requests')
```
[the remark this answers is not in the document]
sam-nguyen7: agree

src/webhook/subscription_validation.py
sam-nguyen7: fixed

PR #25, opened 2020-02-04
OSD-2744 Added webhook validation for updating namespaces

csheremeta:
https://issues.redhat.com/browse/OSD-2744

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ IMAGETAG ?= latest
 CABUNDLECONFIGMAP ?= webhook-cert
 VWC_ANNOTATION ?= managed.openshift.io/inject-cabundle-from
 
-IMG ?= quay.io/lseelye/python3-webhookbase
+IMG ?= quay.io/cshereme/python3-webhookbase
 
 default: all
 all: build-base render
--- a/src/webhook/namespace_validation.py
+++ b/src/webhook/namespace_validation.py
@@ -1,0 +1,56 @@
+from flask import request, Blueprint, Response
+import sys, traceback
+import json
+import os
+import prometheus_client
+import re
+from prometheus_client import Counter
+from prometheus_client.core import CollectorRegistry
+
+from webhook.request_helper import validate, responses
+
+bp = Blueprint("namespace-webhook", __name__)
+
+# define what we track, declare Counter, how many times this route is accessed
+TOTAL_NAMESPACE = Counter('webhook_namespace_validation_total', 'The total number of namespace validation requests')
+DENIED_NAMESPACE = Counter('webhook_namespace_validation_denied', 'The total number of namespace validation requests denied')
+
+@bp.route('/namespace-validation', methods=['POST'])
+def handle_request():
+  # inc total namespace counter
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -9,7 +9,7 @@ IMAGETAG ?= latest
 CABUNDLECONFIGMAP ?= webhook-cert
 VWC_ANNOTATION ?= managed.openshift.io/inject-cabundle-from
 
-IMG ?= quay.io/lseelye/python3-webhookbase
+IMG ?= quay.io/cshereme/python3-webhookbase
```
csheremeta: @lisa where should we be pushing these images?

src/webhook/namespace_validation.py
```diff
@@ -0,0 +1,56 @@
...
+  
+  try:
+    body_dict = request.json['request']
+    requester_group_memberships = body_dict['userInfo']['groups']
+    if "dedicated-admins" in requester_group_memberships:
+      requested_ns = body_dict['namespace']
+      privileged_namespace_re = '(^kube-.*|^openshift.*|^ops-health-monitoring$|^management-infra$|^default$|^logging$|^sre-app-check$|^redhat-.*)'
+      # match will return a match object if the namespace matches the regex, or None if the namespace doesn't match the regex
+      if re.match(privileged_namespace_re, requested_ns) is not None:
+        return responses.response_deny(req=body_dict, msg="You cannot update the privileged namespace {}.".format(requested_ns))
```
csheremeta: @rogbas is this verbiage ok?

Makefile
lisa: @csheremeta I believe the changes made by @robotmaxtron to the build scripts (eg in build/*.sh) handle where the images end up. In many ways the Makefile definition is obsoleted now.

templates/10-namespace-validation.ValidatingWebhookConfiguration.yaml.tmpl
```diff
@@ -0,0 +1,27 @@
...
+    name: namespace-validation.managed.openshift.io
+    rules: 
+      - operations: 
+          - "UPDATE"
+        apiGroups: 
+          - ""
+        apiVersions: 
+          - "*"
+        resources: 
+          - "namespaces"
```
rogbas: Could we please have a Cluster `scope` specified here, just like we have for the group-validation webhook?

src/webhook/namespace_validation.py
rogbas: Verbiage is good @csheremeta thanks!

templates/10-namespace-validation.ValidatingWebhookConfiguration.yaml.tmpl
csheremeta: done!

rogbas: Probably we could get away without it, but standards-wise it's better to keep the same approach.

I checked and the default scope is value "*", which could potentially cause some issues.

PR #28, opened 2020-02-07
Change initcontainer cmd to python3

csheremeta:


lisa:
Addresses issue from #27; the UBI doesn't include a `python` in `$PATH`

cblecker:
Thanks for tracking this down, and sorry for the trouble :)

PR #29, opened 2020-02-18
Fix bug causing exceptions when trying to delete groups OSD-2923

csheremeta:
https://issues.redhat.com/browse/OSD-2923

### DIFF
--- a/src/webhook/group_validation.py
+++ b/src/webhook/group_validation.py
@@ -39,7 +39,11 @@ def handle_request():
 
   try:
     body_dict = request.json['request']
-    group_name = body_dict['object']['metadata']['name']
+    # If trying to delete a group, must get group name from oldObject instead of object
+    if body_dict['object'] is None:
+      group_name = body_dict['oldObject']['metadata']['name']
+    else:
+      group_name = body_dict['object']['metadata']['name']
     userinfo = body_dict['userInfo']
     if userinfo['username'] in ("kube:admin", "system:admin"):
       # kube/system admin can do anything


src/webhook/group_validation.py
```diff
@@ -39,7 +39,11 @@ def handle_request():
 
   try:
     body_dict = request.json['request']
-    group_name = body_dict['object']['metadata']['name']
+    # If trying to delete a group, must get group name from oldObject instead of object
```
lisa: Is this because the `AdmissionReview` for an object being deleted changes the structure that holds information about the subject to `oldObject` (and not in `object)?

csheremeta: that's what it looks like, based on the error message we were seeing in the exception being thrown:
`Exception when trying to access attributes. Request body: {'kind': 'AdmissionReview', 'apiVersion': 'admission.k8s.io/v1beta1', 'request': {'uid': 'c413a5a1-ca1d-4de8-b236-9608b0b5f378', 'kind': {'group': 'user.openshift.io', 'version': 'v1', 'kind': 'Group'}, 'resource': {'group': 'user.openshift.io', 'version': 'v1', 'resource': 'groups'}, 'requestKind': {'group': 'user.openshift.io', 'version': 'v1', 'kind': 'Group'}, 'requestResource': {'group': 'user.openshift.io', 'version': 'v1', 'resource': 'groups'}, 'name': 'test-group1', 'operation': 'DELETE', 'userInfo': {'username': 'cshereme@redhat.com', 'groups': ['dedicated-admins', 'osd-sre-admins', 'osd-sre-cluster-admins', 'system:authenticated:oauth', 'system:authenticated'], 'extra': {'scopes.authorization.openshift.io': ['user:full']}}, 'object': None, 'oldObject': {'kind': 'Group', 'apiVersion': 'user.openshift.io/v1', 'metadata': {'name': 'test-group1', 'uid': 'd7ec5555-26dc-452f-b564-06f47766407b', 'resourceVersion': '34155', 'creationTimestamp': '2020-02-13T16:30:14Z'}, 'users': None}, 'dryRun': False, 'options': {'kind': 'DeleteOptions', 'apiVersion': 'meta.k8s.io/v1', 'propagationPolicy': 'Background'}}}
Backtrace:
------------------------------------------------------------
Traceback (most recent call last):
  File "/app/webhook/group_validation.py", line 42, in handle_request
    group_name = body_dict['object']['metadata']['name']
TypeError: 'NoneType' object is not subscriptable`

lisa: Cool. TIL

/approve
/lgtm

PR #30, opened 2020-03-16
Switch to deployments apps/v1

jewzaam:
https://issues.redhat.com/browse/OSD-3126


lisa:
seems reasonable

/lgtm

PR #31, opened 2020-03-16
Added missing selector to deployment

jewzaam:
https://issues.redhat.com/browse/OSD-3126

### DIFF
--- a/templates/20-validation-webhook.deployment.yaml.tmpl
+++ b/templates/20-validation-webhook.deployment.yaml.tmpl
@@ -8,6 +8,9 @@ metadata:
   namespace: "#NAMESPACE#"
 spec:
   replicas: 3
+  selector:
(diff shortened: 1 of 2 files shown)


templates/20-validation-webhook.deployment.yaml.tmpl
```diff
@@ -8,6 +8,9 @@ metadata:
   namespace: "#NAMESPACE#"
 spec:
   replicas: 3
+  selector:
+    matchLabels:
+      app: validation-webhook
```
rogbas: Just a minor thing, there is a placeholder for that, that gets replaced with the service name: `"#SVCNAME#"`

jewzaam: Updated.

PR #32, opened 2020-03-19
Generic webhook that denies all requests to regular users

jewzaam:
https://issues.redhat.com/browse/OSD-2686

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,16 +11,18 @@ VWC_ANNOTATION ?= managed.openshift.io/inject-cabundle-from
 
 IMG ?= quay.io/cshereme/python3-webhookbase
 
+CONTAINER_ENGINE=$(shell which podman 2>/dev/null || which docker 2>/dev/null)
+
 default: all
 all: build-base render
 
 .PHONY: build-base
 build-base: build/Dockerfile
-	podman build -t $(IMG):$(IMAGETAG) -f build/Dockerfile . 
+	$(CONTAINER_ENGINE) build -t $(IMG):$(IMAGETAG) -f build/Dockerfile . 
 
 .PHONY: push-base
 push-base: build/Dockerfile
-	podman push $(IMG):$(IMAGETAG)
+	$(CONTAINER_ENGINE) push $(IMG):$(IMAGETAG)
 
 # TODO: Change the render to allow for the permissions to have a list of all the webhook names
 # TODO: Pull that list of names from the yaml files?
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,11 @@ all: build-base render
 
 .PHONY: build-base
 build-base: build/Dockerfile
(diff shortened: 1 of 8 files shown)


src/webhook/regular_user_validation.py
```diff
@@ -0,0 +1,55 @@
...
+import prometheus_client
+from prometheus_client import Counter
+from prometheus_client.core import CollectorRegistry
+
+from webhook.request_helper import validate, responses
+
+bp = Blueprint("regular-user-webhook", __name__)
+
+# define what we track, declare Counter, how many times this route is accessed
+REGULAR_USER_DENIED = Counter('webhook_regular_user_denied', 'The total number of regular-user requests denied', labelnames=["request_kind","request_namespace","request_username","request_operation"])
```
lisa: I think this is redundant with Kubernetes 1.16+ (OCP 4.3 is >=1.16), so we should [get them for free](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#monitoring-admission-webhooks)? Maybe we can trim them here?

src/webhook/regular_user_validation.py
```diff
@@ -0,0 +1,55 @@
...
+  except Exception:
+    valid = False
+
+  if not valid:
+    # inc denied counter
+    REGULAR_USER_DENIED.labels(None,None,None,None).inc()
+    return responses.response_invalid()
+  
+  try:
+    # get the username and decide if it's a special user (SA, kube:admin, etc) or not..
```
lisa: `system:unauthenticated` contains a colon, which should be denied by RBAC policy.

Makefile
```diff
@@ -16,7 +16,11 @@ all: build-base render
 
 .PHONY: build-base
 build-base: build/Dockerfile
-	docker build -t $(IMG):$(IMAGETAG) -f build/Dockerfile . 
+	podman build -t $(IMG):$(IMAGETAG) -f build/Dockerfile . 
+
+.PHONY: push-base
+push-base: build/Dockerfile
+	podman push $(IMG):$(IMAGETAG)
```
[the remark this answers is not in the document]
lisa: To clarify: https://github.com/openshift/configure-alertmanager-operator/blob/master/standard.mk#L32 is where that takes place.

Makefile
```diff
@@ -11,12 +11,18 @@ VWC_ANNOTATION ?= managed.openshift.io/inject-cabundle-from
 
 IMG ?= quay.io/cshereme/python3-webhookbase
 
+CONTAINER_ENGINE=$(shell which podman 2>/dev/null || which docker 2>/dev/null)
```
lisa:
```suggestion
CONTAINER_ENGINE ?= $(shell which podman 2>/dev/null || which docker 2>/dev/null)
```

Can let someone override it at the shell if they want.

lisa:
@jewzaam feel free to remove the hold when you're ready to merge.

PR #33, opened 2020-03-20
Fixed generation of selectorsyncset

jewzaam:
Updated so it's generated by default when you build and fixed how it is generated.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ GIT_HASH=$(shell git rev-parse --short=7 HEAD)
 IMAGETAG="${GIT_HASH}"
 CABUNDLECONFIGMAP ?= webhook-cert
 VWC_ANNOTATION ?= managed.openshift.io/inject-cabundle-from
-QUAY_USER=app-sre
+QUAY_USER ?= app-sre
 
 IMG ?= quay.io/${QUAY_USER}/${BASE_IMG}
 
--- a/Makefile
+++ b/Makefile
@@ -2,19 +2,33 @@ SHELL := /usr/bin/env bash
 
 TEMPLATEFILES := $(shell find ./templates -type f -name "*.yaml.tmpl")
 
+BASE_IMG="managed-cluster-validating-webhooks"
 NAMESPACE ?= openshift-validation-webhook
 SVCNAME ?= validation-webhook
 SANAME ?= validation-webhook
-IMAGETAG ?= latest
+GIT_HASH=$(shell git rev-parse --short=7 HEAD)
+IMAGETAG="${GIT_HASH}"
 CABUNDLECONFIGMAP ?= webhook-cert
 VWC_ANNOTATION ?= managed.openshift.io/inject-cabundle-from
+QUAY_USER=app-sre
 
-IMG ?= quay.io/cshereme/python3-webhookbase
+IMG ?= quay.io/${QUAY_USER}/${BASE_IMG}
+
+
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -2,19 +2,33 @@ SHELL := /usr/bin/env bash
...
+BASE_IMG="managed-cluster-validating-webhooks"
 NAMESPACE ?= openshift-validation-webhook
 SVCNAME ?= validation-webhook
 SANAME ?= validation-webhook
-IMAGETAG ?= latest
+GIT_HASH=$(shell git rev-parse --short=7 HEAD)
+IMAGETAG="${GIT_HASH}"
 CABUNDLECONFIGMAP ?= webhook-cert
 VWC_ANNOTATION ?= managed.openshift.io/inject-cabundle-from
+QUAY_USER=app-sre
```
lisa:
```suggestion
QUAY_USER ?= app-sre
```

To allow overriding

Makefile
```diff
@@ -2,19 +2,33 @@ SHELL := /usr/bin/env bash
...
 
-IMG ?= quay.io/cshereme/python3-webhookbase
+IMG ?= quay.io/${QUAY_USER}/${BASE_IMG}
+
+
+SELECTOR_SYNC_SET_TEMPLATE_DIR=deploy
+BUILD_DIRECTORY=build
+SELECTOR_SYNC_SET_DESTINATION=build/selectorsyncset.yaml
+REPO_NAME=managed-cluster-validating-webhooks
```
lisa:
```suggestion
REPO_NAME ?= managed-cluster-validating-webhooks
```

To allow overriding

templates/20-validation-webhook.deployment.yaml.tmpl
```diff
@@ -18,15 +18,15 @@ spec:
     spec:
       serviceAccountName: "#SANAME#"
       initContainers:
-      - image: "#IMG#:#IMAGETAG#"
+      - image: "#IMG#:${IMAGE_TAG}"
```
lisa: Where does `${IMAGE_TAG}` come from? How does it relate to the `${IMAGETAG}` from the Makefile?

Makefile
jewzaam: Why would we allow this to be overriden?

templates/20-validation-webhook.deployment.yaml.tmpl
jewzaam: IMAGE_TAG comes from the template which saasheader applies.  Defined here:
https://github.com/openshift/managed-cluster-validating-webhooks/blob/master/build/00-osd-managed-cluster-validating-webhooks.selectorsyncset.yaml.tmpl#L4-L5

Makefile
lisa: I see it related to the previous suggestion: if I'm doing development locally and for some reason don't have the image in a registry named `managed-cluster-validating-webhooks`, it could be useful to change the name.

jewzaam: Changing the registry  I can see.  Changing the image doesn't make sense.  Testing locally or with my own user I don't need to changing the image name.  I'd just need to push to `nmalik` in quay.io vs pushing to `app-sre` which I don't have permission to.  But I see we permit this in other places so I'll accept it with the expectation we'll *eventually* have centralized standardization for these things.

templates/20-validation-webhook.deployment.yaml.tmpl
lisa: Ahh, I see. Thanks!

PR #37, opened 2020-03-26
Reverting to old way of pushing image in the bash script.

jewzaam:
The move to make isn't working.
Debugging requires a PR and merge, which is a pain.
This moves back to what it was doing before in the bash script.
If it doesn't work I'll have to install skopeo to reproduce locally but hoping not.


jewzaam:
/hold

prefer https://github.com/openshift/managed-cluster-validating-webhooks/pull/38 if it fixes the issues

jewzaam:
Got things fixed in https://github.com/openshift/managed-cluster-validating-webhooks/pull/38

PR #39, opened 2020-03-26
QUAY_{USER,TOKEN} provided by env vars

jewzaam:


jewzaam:
Merging.  It's broken already so I can't make it worse!

Makefile
```diff
@@ -10,9 +10,8 @@ GIT_HASH=$(shell git rev-parse --short=7 HEAD)
 IMAGETAG=${GIT_HASH}
 CABUNDLECONFIGMAP ?= webhook-cert
 VWC_ANNOTATION ?= managed.openshift.io/inject-cabundle-from
-QUAY_USER ?= app-sre
 
-IMG ?= quay.io/${QUAY_USER}/${BASE_IMG}
+IMG ?= quay.io/app-sre/${BASE_IMG}
```
lisa: I don't think this is needed. Line 13, `QUAY_USER ?= app-sre` only sets the `QUAY_USER` variable if it us otherwise unset, thus, if I'm doing development work and need to test with a `QUAY_USER=lseelye` I can set that in the environment variables, but at prod time, this will not work.

jewzaam:
FWIW the issue is the QUAY_USER value is set by scripts in cicd and the value set in Makefile is not what is expected hence skopeo fails to auth.  This change allows the env var to be used and updates IMG to what we expect by default.  Local testing can override IMG

PR #40, opened 2020-04-03
Fix Bug: Regular-User validation webhook is treating SREs as regular users

jewzaam:
https://issues.redhat.com/browse/OSD-3242

SRE is now allowed to modify the resources the regular user validation webhook gates.
Added unit tests for regular_user_validation (yay!)
Added generation of requirements.txt (just in case you need it)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,6 @@ VWC_ANNOTATION ?= managed.openshift.io/inject-cabundle-from
 
 IMG ?= quay.io/app-sre/${BASE_IMG}
 
-
 SELECTOR_SYNC_SET_TEMPLATE_DIR=deploy
 BUILD_DIRECTORY=build
 SELECTOR_SYNC_SET_DESTINATION=build/selectorsyncset.yaml
@@ -24,14 +23,22 @@ CONTAINER_ENGINE?=docker
 default: all
 all: build-base build-sss
 
+.PHONY: clean
+clean:
+	$(CONTAINER_ENGINE) rmi $(REPO_NAME):test || true
+	$(CONTAINER_ENGINE) rmi $(IMG):$(IMAGETAG) || true
+
+.PHONY: test-container
+test-container:
+	$(CONTAINER_ENGINE) build -t $(REPO_NAME):test -f build/Dockerfile.test .
+
 .PHONY: test
-test:
-	python -m unittest discover src -vvv
+test: test-container
+	$(CONTAINER_ENGINE) run --rm -v `pwd -P`:`pwd -P` $(REPO_NAME):test /bin/sh -c "cd `pwd`; python -m unittest discover src -vvv"
 
 .PHONY: build-sss
 build-sss: render
 	${CONTAINER_ENGINE} run --rm -v `pwd -P`:`pwd -P` python:2.7.15 /bin/sh -c "cd `pwd`; pip install oyaml; python build/generate_syncset.py -t ${SELECTOR_SYNC_SET_TEMPLATE_DIR} -b ${BUILD_DIRECTORY} -d ${SELECTOR_SYNC_SET_DESTINATION} -r ${REPO_NAME}"
-
 
 .PHONY: build-base
 build-base: test build/Dockerfile
@@ -51,7 +58,6 @@ skopeo-push:
 		"docker-daemon:${IMG}:${IMAGETAG}" \
 		"docker://${IMG}:${IMAGETAG}"
 
-
 # TODO: Change the render to allow for the permissions to have a list of all the webhook names
 # TODO: Pull that list of names from the yaml files?
 render: $(TEMPLATEFILES) build/Dockerfile
@@ -70,8 +76,9 @@ render: $(TEMPLATEFILES) build/Dockerfile
 .PHONY: requirements
 requirements:
 	if [ "$(pip list | grep pipreqs | wc -l)" != "0" ]; then \
-		rm -f requirements.txt; \
+		rm -f src/requirements.txt; \
 		pipreqs ./; \
+		mv requirements.txt src/; \
 	else \
 		echo "FAILURE please install pipreqs: pip install pipreqs"; \
 	fi
--- a/Makefile
+++ b/Makefile
@@ -24,13 +24,17 @@ CONTAINER_ENGINE?=docker
 default: all
 all: build-base build-sss
 
+.PHONY: test
+test:
+	python -m unittest discover src -vvv
+
 .PHONY: build-sss
 build-sss: render
 	${CONTAINER_ENGINE} run --rm -v `pwd -P`:`pwd -P` python:2.7.15 /bin/sh -c "cd `pwd`; pip install oyaml; python build/generate_syncset.py -t ${SELECTOR_SYNC_SET_TEMPLATE_DIR} -b ${BUILD_DIRECTORY} -d ${SELECTOR_SYNC_SET_DESTINATION} -r ${REPO_NAME}"
 
 
 .PHONY: build-base
-build-base: build/Dockerfile
+build-base: test build/Dockerfile
 	$(CONTAINER_ENGINE) build -t $(IMG):$(IMAGETAG) -f build/Dockerfile . 
 
 .PHONY: push-base
@@ -62,3 +66,13 @@ render: $(TEMPLATEFILES) build/Dockerfile
 			-e "s!\#VWC_ANNOTATION\#!$(VWC_ANNOTATION)!g" \
 		$$f > deploy/$$(basename $$f .tmpl) ;\
 	done
+
+.PHONY: requirements
+requirements:
+	if [ "$(pip list | grep pipreqs | wc -l)" != "0" ]; then \
+		rm -f requirements.txt; \
+		pipreqs ./; \
+	else \
+		echo "FAILURE please install pipreqs: pip install pipreqs"; \
+	fi
+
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -24,13 +23,25 @@ CONTAINER_ENGINE?=docker
 default: all
 all: build-base build-sss
 
+.PHONY: clean
+clean:
+	$(CONTAINER_ENGINE) rmi $(REPO_NAME):test || true
+	$(CONTAINER_ENGINE) rmi $(IMG):$(IMAGETAG) || true
```
lisa, line 29:
```suggestion
	$(CONTAINER_ENGINE) rmi $(REPO_NAME):test $(IMG):$(IMAGETAG) || true
```

Makefile
```diff
@@ -24,13 +23,25 @@ CONTAINER_ENGINE?=docker
 default: all
 all: build-base build-sss
```
lisa, line 24: Is it worth looping in the `test` target here?

src/webhook/regular_user_validation.py
```diff
@@ -25,22 +31,33 @@ def handle_request():
...
-    return responses.response_invalid()
\ No newline at end of file
+    return responses.response_invalid()
+
+
+def is_request_allowed(username, groupname, admin_groupnames=[]):
+  if groupname not in admin_groupnames and (username == "system:unauthenticated" or (not username.startswith("kube:") and not username.startswith("system:"))):
+    return False
+  else:
+    return True
```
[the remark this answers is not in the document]
jewzaam: Probably could.  My initial try didn't work so I just did this to get it done..

Makefile
jewzaam: Torn on this, I like using something different simply because it is going to be harder to accidentally push since it's not setup for that.  It isn't something we should be thinking we can push.. thoughts?

Makefile
jewzaam: It's pulled in as a dependency by build-base.

src/webhook/regular_user_validation.py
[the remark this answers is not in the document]
2uasimojo: IMHO it would be more scrutable if you split it up _further_.
```
if groupname in admin_groupnames:
    return True
if username == "system:authenticated":
    return True
if username.split(":", 1)[0] in ("kube", "system"):
    return True
return False
```

src/webhook/regular_user_validation.py
```diff
@@ -25,22 +31,33 @@ def handle_request():
...
+
+    if body_dict['object'] is None:
+      group_name = body_dict['oldObject']['metadata']['name']
+    else:
+      group_name = body_dict['object']['metadata']['name']
+
     # get the username and decide if it's a special user (SA, kube:admin, etc) or not.. 
     # deny any request if it's not a special user:  kube:*, system:* except system:unauthenticated
+    # ALLOW any request if the group_name is in admin_groups
     # reference: https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#request
```
2uasimojo, line 45: Consider moving this comment into `is_request_allowed` itself. (Perhaps even as a docstring.)

src/webhook/regular_user_validation.py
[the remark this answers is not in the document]
jewzaam: Pushed an update to `return not(<the if condition>)`.
Suggestions welcome, but I think it's reasonable to be able to read it as it is.

src/webhook/regular_user_validation.py
jewzaam: I'm weak on my python-foo, suggestions for the changes?

src/webhook/regular_user_validation.py
```diff
@@ -25,22 +31,30 @@ def handle_request():
...
-    body_dict = request.json['request']
-    username = body_dict['userInfo']['username']
-    if username == "system:unauthenticated" or (not username.startswith("kube:") and not username.startswith("system:")):
+    if is_request_allowed(username, group_name, admin_groups):
```
2uasimojo, line 46: Wait wait, isn't this backwards now?
```
if is_request_allowed():
    return responses.response_deny()
else:
    return responses.response_allow()
```

src/webhook/regular_user_validation.py
```diff
@@ -25,22 +31,30 @@ def handle_request():
...
+
+    if body_dict['object'] is None:
+      group_name = body_dict['oldObject']['metadata']['name']
+    else:
+      group_name = body_dict['object']['metadata']['name']
+
     # get the username and decide if it's a special user (SA, kube:admin, etc) or not.. 
     # deny any request if it's not a special user:  kube:*, system:* except system:unauthenticated
+    # ALLOW any request if the group_name is in admin_groups
     # reference: https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#request
```
2uasimojo, line 45: Suggestion part 1. The details of how you're deciding whether the request should be allowed are now within `is_request_allowed` and don't need to be here.
```suggestion
```

src/webhook/regular_user_validation.py
```diff
@@ -25,22 +31,30 @@ def handle_request():
...
   except Exception:
     print("Exception:")
     print("-"*60)
     traceback.print_exc(file=sys.stdout)
-    return responses.response_invalid()
\ No newline at end of file
+    return responses.response_invalid()
+
+
+def is_request_allowed(username, groupname, admin_groupnames=[]):
```
2uasimojo, line 59: Suggestion part 2, use a docstring for the method and include the reference.
```suggestion
def is_request_allowed(username, groupname, admin_groupnames=()):
    """Decide if it's a special user (SA, kube:admin, etc) or not.

    reference: https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#request
    """
```

src/webhook/regular_user_validation.py
```diff
@@ -25,22 +31,30 @@ def handle_request():
...
+  return not(groupname not in admin_groupnames and (username == "system:unauthenticated" or (not username.startswith("kube:") and not username.startswith("system:"))))
```
2uasimojo, line 60: Yeah, this is way hard to understand IMO, and very not-pythonic.
Suggestion part 3:
```suggestion
  # Explicitly unauthenticated
  if username == "system:unauthenticated":
      return False
  if groupname in admin_groupnames:
      return True
  if username.startswith("kube:") or username.startswith("system:"):
      return True
  # Deny anyone else
  return False
```
If you do this, you don't need the comments removed from above at all:
```
    # deny any request if it's not a special user:  kube:*, system:* except system:unauthenticated
    # ALLOW any request if the group_name is in admin_groups
```
...because that's obvious from the code.

src/webhook/regular_user_validation.py
2uasimojo: See below.

src/webhook/regular_user_validation.py
jewzaam: Yup, pushed a fix with unit tests.  Thanks!

Makefile
lisa: > Torn on this, I like using something different simply because it is going to be harder to accidentally push since it's not setup for that. It isn't something we should be thinking we can push.. thoughts?

I'm not sure I follow. This is the `clean` target, where we want to remove the exiting images. The two `rmi` can be condensed into a single one. Am I missing something?

jewzaam: ack, somehow I missed what the change was.. merging thanks

jewzaam:
Think all things are resolved now.  Refactored request handling to a separate function and added unit tests.  It did indeed fail the tests until logic was fixed.  Thanks @2uasimojo !

src/webhook/test_regular_user_validation.py
```diff
@@ -1,23 +1,70 @@
...
+                "object": {
+                    "kind": kind,
+                    "metadata": {
+                        "name": groupname
+                    }
+                },
+            }
+        }
+    }
+    return request
```
2uasimojo, line 34: Following on from above, I think you could avoid having to change the way you're accessing `request.json` by making this:

```suggestion
    class FakeRequest(object):
        json = {
            "request": {
                "uid": "testuser",
                "operation": "UPDATE",
                "userInfo": {
                    "username": username
                },
                "oldObject": {
                    "metadata": {
                        "name": groupname
                    }
                },
                "object": {
                    "kind": kind,
                    "metadata": {
                        "name": groupname
                    }
                },
            }
        }

    return FakeRequest()
```

src/webhook/test_regular_user_validation.py
```diff
@@ -1,5 +1,6 @@
 import unittest
 import json
+import enum
```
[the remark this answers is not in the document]
2uasimojo: Confirmed...
```
[efried@efried managed-cluster-validating-webhooks]$ flake8 --select=F401 src/webhook/test_regular_user_validation.py 
src/webhook/test_regular_user_validation.py:3:1: F401 'enum' imported but unused
```
...but this isn't the only pep8 error in this file
```
[efried@efried managed-cluster-validating-webhooks]$ flake8 src/webhook/test_regular_user_validation.py  | wc -l
11
```
or in this repo
```
[efried@efried managed-cluster-validating-webhooks]$ find . -name '*.py' | xargs flake8 | wc -l
242
```
so probably not something to attack in this PR.

Could consider a future PR to add flake8 to the CI though...

2uasimojo:
I'd feel better if @lisa gave this a nod as well before it was merged, but
/lgtm

PR #42, opened 2020-04-06
Fixing where group info is pulled for 'regular user' webhook

jewzaam:
https://issues.redhat.com/browse/OSD-3242

Original logic was not looking at the incoming user's groups, as expressed in userInfo.
All SRE requests were still being denied because it didn't have any group info.  This fixes it.


src/webhook/regular_user_validation.py
```diff
@@ -57,17 +53,20 @@ def get_response(req, debug=False):
...
-  if groupname in admin_groupnames:
-      return True
+    return False
+  # system users is ALLOW
   if username.startswith("kube:") or username.startswith("system:"):
+    return True
+  # SRE groups are ALLOW
+  for group in groups:
+    if group in admin_groupnames:
       return True
```
2uasimojo, line 70: Another way to do this that's O(N) would be
```suggestion
  if set(groups) & set(admin_groupnames):
      return True
```
but I'm guessing both lists are so small this will actually wind up being _less_ efficient...

src/webhook/test_regular_user_validation.py
```diff
@@ -9,27 +9,22 @@
...
                     "userInfo": {
-                        "username": username
-                    },
-                    "oldObject": {
-                        "metadata": {
-                            "name": groupname
-                        }
+                        "username": username,
+                        "groups": [
+                            groupname
```
2uasimojo, line 22: Really should provide a way to let this contain multiple group names, and tests that do that.

src/webhook/test_regular_user_validation.py
```diff
@@ -40,34 +35,34 @@ def __getitem__(self, item):
...
-        self.assertTrue(is_request_allowed("kube:admin", "", ADMIN_GROUPS))
+        self.assertTrue(is_request_allowed("kube:admin", [], ADMIN_GROUPS))
 
     def test_is_request_allowed_systemadmin(self):
-        self.assertTrue(is_request_allowed("system:admin", "", ADMIN_GROUPS))
+        self.assertTrue(is_request_allowed("system:admin", [], ADMIN_GROUPS))
 
     def test_is_request_allowed_nonadmin_group(self):
-        self.assertFalse(is_request_allowed("someuser", "some-group", ADMIN_GROUPS))
+        self.assertFalse(is_request_allowed("someuser", ["some-group"], ADMIN_GROUPS))
```
2uasimojo, line 50: Missing tests with variations on `admin_groupnames`, including permutations where `groupnames` contains groups that would normally be allowed under `ADMIN_GROUPS`. Also permutations proving that `system:*`/`kube:*` is allowed because it's hitting that code path rather than the one that passes based on groups. (I should have caught that in PR #40, sorry.)

src/webhook/regular_user_validation.py
2uasimojo: In case unclear, feel free to disregard this one.

jewzaam: I like not having have an if in the loop.  I thought I'd seen some for logic somewhere that had if conditions but couldn't find it quickly.  I'm OK with the change just for that, but if performance is worse then will skip.  Yes they are small arrays.

2uasimojo: You're probably thinking of list comprehensions, something like
`[group for group in groups if group in admin_groupnames]`.
This produces a list. In this context, if the list was nonempty, it should trigger the condition.
It's close to the same thing you've got -- except that it generates the *whole* list rather than stopping on the first hit. Same thing with the `set` solution I suggested; the difference is that it only has to run through `admin_groupnames` once vs `len(groups)` times.

Anyway, this is all totally academic. What you've got is fine. (My holdup on this PR is for test coverage.)

src/webhook/test_regular_user_validation.py
jewzaam: Latest commit should cover more permutations, thanks.

cblecker:
/approve

This looks good to me.

@2uasimojo: You want to give this one more look over? If it looks good to you, feel free to `/lgtm` and let this merge.

2uasimojo:
Thanks for expanding the test coverage.

There's still not much of a nod to variations in the `admin_groupnames`, but I get the sense those don't change a whole lot.

Much, much deduplication could be accomplished via subclassing (simple) and/or [ddt](https://ddt.readthedocs.io/en/latest/) (moar better), but that can be worked on later.

/lgtm

2uasimojo:
> Much, much deduplication could be accomplished via subclassing (simple) and/or [ddt](https://ddt.readthedocs.io/en/latest/) (moar better), but that can be worked on later.

I proposed this (the subclassing route) via PR #49.

PR #47, opened 2020-04-08
Un-ignore and fix remaining pep8 rules

2uasimojo:
Commit 72ef319 via PR #41 added a flake8 check for python files in this
repo, but ignored several of the rules for the sake of expediency. This
commit pays down that technical debt by removing the ignores and fixing
those remaining violations.

There is no functional change. All the python file deltas are pure
formatting *except*:
- The privileged_namespace_re in src/webhook/namespace_validation.py is
extracted, simplified, and pre-compiled.
- Imports are tweaked to comply with best practices (import only
modules; sort imports alphabetically; group by system, 3rd party,
local).
- A couple of variables are extracted for readability and efficiency.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,14 +33,7 @@ test-container:
 
 .PHONY: lint
 lint: test-container
-	# E111 indentation is not a multiple of four
-	# E121 continuation line under-indented for hanging indent
-	# E401 multiple imports on one line
-	# E402 module level import not at top of file
-	# E501 line too long (N > 79 characters)
-	# E722 do not use bare 'except'
-	# W293 blank line contains whitespace
-	$(CONTAINER_ENGINE) run --rm -v `pwd -P`:`pwd -P` $(REPO_NAME):test /bin/sh -c "cd `pwd`; flake8 --ignore E111,E121,E114,E401,E402,E501,E722,W293 src/"
+	$(CONTAINER_ENGINE) run --rm -v `pwd -P`:`pwd -P` $(REPO_NAME):test /bin/sh -c "cd `pwd`; flake8 src/ build/"
 
 .PHONY: test
 test: test-container
--- a/build/generate_syncset.py
+++ b/build/generate_syncset.py
@@ -1,30 +1,36 @@
 #!/usr/bin/env python
 
-import oyaml as yaml
-
 import os
-import sys
 import argparse
 import copy
 
+import oyaml as yaml
+
+
+TPL = "00-osd-managed-cluster-validating-webhooks.selectorsyncset.yaml.tmpl"
+
+
 def get_yaml_all(filename):
-    with open(filename,'r') as input_file:
+    with open(filename, 'r') as input_file:
         return list(yaml.load_all(input_file))
 
+
 def get_yaml(filename):
-    with open(filename,'r') as input_file:
+    with open(filename, 'r') as input_file:
         return yaml.safe_load(input_file)
+
 
 def get_all_yaml_files(path):
     file_paths = []
-    for r,d,f in os.walk(path):
+    for r, _, f in os.walk(path):
         for file in f:
             if file.endswith('.tmpl') or file.endswith('.yaml'):
-                file_paths.append(os.path.join(r,file))
+                file_paths.append(os.path.join(r, file))
         # break, so we don't recurse
         break
     file_paths = sorted(file_paths)
     return file_paths
+
 
 def get_all_yaml_obj(file_paths):
     yaml_objs = []
@@ -34,37 +40,43 @@ def get_all_yaml_obj(file_paths):
             yaml_objs.append(obj)
     return yaml_objs
 
+
 def process_yamls(directory):
     # Get all yaml files as array of yaml objects
     yamls = get_all_yaml_obj(get_all_yaml_files(directory))
     if len(yamls) == 0:
         return
-    
-    
 
-    
     for obj in template_data['objects']:
         obj['spec']['resources'] = []
         if obj['kind'] == 'SelectorSyncSet':
             for y in yamls:
                 obj['spec']['resources'].append(y)
-    
+
     output_data['objects'].append(obj)
-    
 
 
 if __name__ == '__main__':
-    #Argument parser
-    parser = argparse.ArgumentParser(description="selectorsyncset generation tool", usage='%(prog)s [options]')
-    parser.add_argument("--template-dir", "-t", required=True, help="Path to template directory [required]")
-    parser.add_argument("--build-dir", "-b", required=True, help="Path to folder containing yaml files [required]")
-    parser.add_argument("--destination", "-d", required=True, help="Destination for selectorsynceset file [required]")
-    parser.add_argument("--repo-name", "-r", required=True, help="Name of the repository [required]")
+    # Argument parser
+    parser = argparse.ArgumentParser(
+        description="selectorsyncset generation tool",
+        usage='%(prog)s [options]')
+    parser.add_argument(
+        "--template-dir", "-t", required=True,
+        help="Path to template directory [required]")
+    parser.add_argument(
+        "--build-dir", "-b", required=True,
+        help="Path to folder containing yaml files [required]")
+    parser.add_argument(
+        "--destination", "-d", required=True,
+        help="Destination for selectorsynceset file [required]")
+    parser.add_argument(
+        "--repo-name", "-r", required=True,
+        help="Name of the repository [required]")
     arguments = parser.parse_args()
 
     # Get the template data
-    template_data = get_yaml(os.path.join(arguments.build_dir, "00-osd-managed-cluster-validating-webhooks.selectorsyncset.yaml.tmpl"))
-    #selectorsyncset_data = get_yaml(os.path.join(arguments.template_dir, "selectorsyncset.yaml"))
+    template_data = get_yaml(os.path.join(arguments.build_dir, TPL))
 
     # The templates and script are shared across repos (copy & paste).
     # Set the REPO_NAME parameter.
@@ -75,10 +87,10 @@ if __name__ == '__main__':
     output_data = copy.deepcopy(template_data)
     output_data.pop("objects")
     output_data['objects'] = []
-    
+
     # for each subdir of yaml_directory append 'object' to template
     process_yamls(arguments.template_dir)
 
     # write template file ordering by keys
-    with open(arguments.destination,'w') as outfile:
+    with open(arguments.destination, 'w') as outfile:
         yaml.dump(output_data, outfile)
--- a/src/gunicorn.py
+++ b/src/gunicorn.py
@@ -1,6 +1,7 @@
 # gunicorn config file
 
-access_log_format = '%(h)s %(l)s %(u)s %(t)s "%(r)s" %(s)s %(b)s "%(f)s" "%(a)s" "pid=%(p)s"'
+access_log_format = (
+    '%(h)s %(l)s %(u)s %(t)s "%(r)s" %(s)s %(b)s "%(f)s" "%(a)s" "pid=%(p)s"')
 raw_env = [
   'FLASK_APP=webhook'
 ]
--- a/src/init.py
+++ b/src/init.py
@@ -1,15 +1,24 @@
 #!/usr/bin/env python3
 
-# Update the ValidatingWebhookConfiguration with the contents of the Service CA.
+# Update the ValidatingWebhookConfiguration with the contents of the Service
+# CA.
 
-from kubernetes import client, config
-import os
 import argparse
-import copy
 import base64
+import copy
+import os
+
+from kubernetes import client
+from kubernetes import config
 
 parser = argparse.ArgumentParser(description="Options to Program")
-parser.add_argument('-a', default="managed.openshift.io/inject-cabundle-from", dest='annotation_name', help='What is the annotation that has a reference to a namespace/configmap for the caBundle. The cert must be stored in pem format in a key called service-ca.crt')
+parser.add_argument(
+    '-a',
+    default="managed.openshift.io/inject-cabundle-from",
+    dest='annotation_name',
+    help='What is the annotation that has a reference to a '
+         'namespace/configmap for the caBundle. The cert must be stored in '
+         'pem format in a key called service-ca.crt')
 parsed = parser.parse_args()
 
 config.load_incluster_config()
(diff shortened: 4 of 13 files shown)


2uasimojo:
From PR #51:

> I'm not sure the code changes for the linter's benefit are worth making the code harder to read for humans.

This is something of a... religious issue :) But objectively:
- Out of curiosity, what are you finding harder to read about the new version? I'll never argue that pep8's rules are perfect for everyone, but...
- Like gofmt (which I saw in at least one other repo of ours) one major benefit is to standardize the formatting not just within a project, but across projects. This generally has the effect of _improving_ readability for anyone accustomed to the standard. Recognizing of course that...
- Like gofmt, one size surely doesn't fit all (there will be some rules that some people find objectionable) but it represents a kind of "least common denominator" that should at least not be horrible for anyone. But to that point...
- Because each language's linter was made for its language and by developers/proponents of its language, transitioning between lint-clean files in e.g. go and python could be jarring if some of the baseline rules (or lack thereof) (e.g. indent width, line length) differ. Is it possible that that's what's hitting you hardest here? Whoever wrote these files was primarily a go developer and followed gofmt-esque rules, and because you're a go developer, that reads naturally for you, whereas pep8 is different enough to be uncomfortable at first?

I'm not going to die on a hill for this change if there's general objection, but I'd like to keep the discussion open for at least a little bit.

jharrington22:
@2uasimojo @lisa this makes sense to me and looks good. I won't label it I'll leave that for you to ok and merge @lisa

lisa:
> From PR #51:
> 
> @lisa said:
> 
> > I'm not sure the code changes for the linter's benefit are worth making the code harder to read for humans.
> 
> This is something of a... religious issue :) But objectively:
> 
> * Out of curiosity, what are you finding harder to read about the new version? I'll never argue that pep8's rules are perfect for everyone, but...
> * Like gofmt (which I saw in at least one other repo of ours) one major benefit is to standardize the formatting not just within a project, but across projects. This generally has the effect of _improving_ readability for anyone accustomed to the standard. Recognizing of course that...
> * Like gofmt, one size surely doesn't fit all (there will be some rules that some people find objectionable) but it represents a kind of "least common denominator" that should at least not be horrible for anyone. But to that point...
> * Because each language's linter was made for its language and by developers/proponents of its language, transitioning between lint-clean files in e.g. go and python could be jarring if some of the baseline rules (or lack thereof) (e.g. indent width, line length) differ. Is it possible that that's what's hitting you hardest here? Whoever wrote these files was primarily a go developer and followed gofmt-esque rules, and because you're a go developer, that reads naturally for you, whereas pep8 is different enough to be uncomfortable at first?
> 
> I'm not going to die on a hill for this change if there's general objection, but I'd like to keep the discussion open for at least a little bit.

I totally get that this is a contentious issue. :) To be up front, my focus is on human readability (computers will deal with it) in most cases. The main issue is the column limitation to ~80 characters. I think it made sense 35 years ago when we had 80 column displays but today we have much better displays that can show hundreds of columns if we want, we don't have to force ourselves into a "small shoe" so to speak.

In some cases it has lead to improvement [ArgumentParser](https://github.com/openshift/managed-cluster-validating-webhooks/pull/47/files#diff-f402a7d9c63332f49145efff62795ddcR14-R21) ("Example 1") and [Prometheus metrics](https://github.com/openshift/managed-cluster-validating-webhooks/pull/47/files#diff-14e35e357456e04373f97b346ce76231R13-R19) ("Example 5"). But I think in most other cases we have a regression in human readability; it is my thesis that these cases are not worth the 80 column constraint.

    # ex 1
    parser = argparse.ArgumentParser(description="Options to Program")
    parser.add_argument('-a', default="managed.openshift.io/inject-cabundle-from", dest='annotation_name', help='What is the annotation that has a reference to a namespace/configmap for the caBundle. The cert must be stored in pem format in a key called service-ca.crt')
    parser.add_argument(
        '-a',
        default="managed.openshift.io/inject-cabundle-from",
        dest='annotation_name',
        help='What is the annotation that has a reference to a '
             'namespace/configmap for the caBundle. The cert must be stored in '
             'pem format in a key called service-ca.crt')
    
    # ex 5
    # define what we track, declare Counter, how many times this route is accessed
    TOTAL_GROUP = Counter('webhook_group_validation_total', 'The total number of group validation requests')
    DENIED_GROUP = Counter('webhook_group_validation_denied', 'The total number of group validation requests denied')
    TOTAL_GROUP = prometheus_client.Counter(
        'webhook_group_validation_total',
        'The total number of group validation requests')
    DENIED_GROUP = prometheus_client.Counter(
        'webhook_group_validation_denied',
        'The total number of group validation requests denied')



[Example 2](https://github.com/openshift/managed-cluster-validating-webhooks/pull/47/files#diff-f402a7d9c63332f49145efff62795ddcR44-R51)

    def get_validating_webhook_configuration_objects_with_annotation(client,
                                                                     annotation):
        ret = []
        for o in client.list_validating_webhook_configuration().items:
            if (o.metadata.annotations is not None and
                    annotation in o.metadata.annotations):
                ret.append(o)
        return ret

In this example, I find the method arguments to be clear and net neutral with the previous code. However, in the `if` statement beginning (now) on line 48 there is a line break split on `is not None and`. When reading this the mind wants to continue reading to the right to the end of line, especially since the previous line extends 16 columns further, so the newline appears out of place (to me). As it happens, If the `if` statement were continued on one line it would only extend to 90 columns. Here, I think it isn't worth conserving 10 columns.

[Example 3](https://github.com/openshift/managed-cluster-validating-webhooks/pull/47/files#diff-f402a7d9c63332f49145efff62795ddcR59-R61) and [Example 4](https://github.com/openshift/managed-cluster-validating-webhooks/pull/47/files#diff-f402a7d9c63332f49145efff62795ddcR69-R73)


        # ex 3
        print("WARNING: Skipping validatingwebhookconfiguration/{}: Couldn't "
              "find a cert from {}/{} ConfigMap. \n".format(
                  vwc.metadata.name, ns, cm_name))

        # ex 4
            print("validatingwebhookconfiguration/{}: Injecting caBundle from "
                  "{}/{}, for hook name {}, to service/{}/{}\n".format(
                      new_vwc.metadata.name, ns, cm_name, hook.name,
                      hook.client_config.service.namespace,
                      hook.client_config.service.name))


In these two examples a multi-line string is being split up. [Example 4](https://github.com/openshift/managed-cluster-validating-webhooks/pull/47/files#diff-f402a7d9c63332f49145efff62795ddcR69-R73) is the more visually challenging of the two because in other languges there's with a `printf`-style method the end of the string (eg when a `"` is seen) it means the next tokens are going to be variables to interpolate into the string. Yes, in Python it is different. In Python I would contend that it is also challenging to keep mental tally of the tokens in the `print`.

I think we should have a column constraint since 50,000 columns is clearly going to be a challenge for human readability too, so what should that be? 150? 200? The median editor size would be a good informant for it. Using a tabstop of 8 I feel contributes to the issue because so much of the screen width is sacrificed to white space (a Python pet peeve of mine).

You wrote also:

> Because each language's linter was made for its language and by developers/proponents of its language, transitioning between lint-clean files in e.g. go and python could be jarring if some of the baseline rules (or lack thereof) (e.g. indent width, line length) differ. Is it possible that that's what's hitting you hardest here? Whoever wrote these files was primarily a go developer and followed gofmt-esque rules, and because you're a go developer, that reads naturally for you, whereas pep8 is different enough to be uncomfortable at first?

Yes, guilty as charged. ;-) I wrote a good portion of the code impacted by this PR. I would have called myself a Ruby developer before a Go developer, so I guess that's a mission accomplished moment for learning Go. It is definitely jarring to see the changes, certainly the column constraints, as I mentioned above.

A separate issue: Do we have a `.vimrc` vscode `settings.json` file (and etc) we can give to folks to help the developer-side authorship?

2uasimojo:
Thank you for the thorough response @lisa.

> The main issue is the column limitation to ~80 characters. I think it made sense 35 years ago when we had 80 column displays but today we have much better displays that can show hundreds of columns if we want, we don't have to force ourselves into a "small shoe" so to speak.

Personally, I happen to agree that 80c is excessively limiting. (Side note [1].) (Side note [2].) (Side note [3].) And I completely agree with all of your examples. So it's my instinct to say that we should just make an exception for this rule and make the limit something more "reasonable". But:
- It's a slippery slope. If we start cherry-picking exceptions for people's pet peeves (even if they're shared among multiple devs who happen to be present at a given time), where does it end?
- What's "reasonable"? If we started negotiating line length, you suggested "150? 200?" I might go as high as 100, but no way would I go longer. Religious war ensues, the discussion quickly becomes counterproductive, and substantive changes get stuck for nontechnical reasons.
- There's still value in "global" consistency: knowing I can take a python file from _anywhere_ and a) not have to resize my editor, change / work around my tabstop/margin settings, etc.; b) port (chunks of) it to another project -- which is obviously a thing in the open source world -- without having to reformat; c) just be able to read the thing because I'm used to the way it looks.

For a few of your specific issues (which, again, I agree with) we can just make different reformatting choices:

> [Example 3](https://github.com/openshift/managed-cluster-validating-webhooks/pull/47/files#diff-f402a7d9c63332f49145efff62795ddcR59-R61) and [Example 4](https://github.com/openshift/managed-cluster-validating-webhooks/pull/47/files#diff-f402a7d9c63332f49145efff62795ddcR69-R73)
> 
> ```
>     # ex 3
>     print("WARNING: Skipping validatingwebhookconfiguration/{}: Couldn't "
>           "find a cert from {}/{} ConfigMap. \n".format(
>               vwc.metadata.name, ns, cm_name))
> 
>     # ex 4
>         print("validatingwebhookconfiguration/{}: Injecting caBundle from "
>               "{}/{}, for hook name {}, to service/{}/{}\n".format(
>                   new_vwc.metadata.name, ns, cm_name, hook.name,
>                   hook.client_config.service.namespace,
>                   hook.client_config.service.name))
> ```
> 
> In these two examples a multi-line string is being split up. [Example 4](https://github.com/openshift/managed-cluster-validating-webhooks/pull/47/files#diff-f402a7d9c63332f49145efff62795ddcR69-R73) is the more visually challenging of the two because in other languges there's with a `printf`-style method the end of the string (eg when a `"` is seen) it means the next tokens are going to be variables to interpolate into the string. 

Splitting up long strings is something you have to deal with in any language. No matter what line length limit (side note [4]) you choose, you're going to have strings that exceed it and will need to be split up in the source. In python, you can do this with extra syntactic sugar (pretty sure you could do it with `+` like [this go example](https://github.com/openshift-online/ocm-cli/blob/master/cmd/ocm/login/cmd.go#L152-L159)) but within parens you're allowed to do it as above, just closing the quotes and reopening them on the next line, which IMO is the easiest to do and the least visually disruptive. Just a matter of getting used to it.

> Yes, in Python it is different. In Python I would contend that it is also challenging to keep mental tally of the tokens in the `print`.

There's an answer for this in all of the interpolation mechanisms in python. For format strings, you can put a key in the brackets, and then use kwargs in your `.format()` (in any order) to match them up. So this example could be:

```
         print("validatingwebhookconfiguration/{vwcname}: Injecting caBundle from "
               "{namespace}/{cmname}, for hook name {hookname}, to "
               "service/{svcnamespace}/{svcname}\n".format(
                   vwcname=new_vwc.metadata.name,
                   namespace=ns,
                   cmname=cm_name,
                   hookname=hook.name,
                   svcnamespace=hook.client_config.service.namespace,
                   svcname=hook.client_config.service.name))
```

But python also supports printf-style interpolations. You can use them like in C (`%s`, `%i`, etc.) but you can also key them like above and use a dict for the substitutions. So the above would look like:

```
         print("validatingwebhookconfiguration/%(vwcname)s: Injecting caBundle from "
               "%(namespace)s/%(cmname)s, for hook name %(hookname)s, to "
               "service/%(svcnamespace)s/%(svcname)s\n" % {
                   'vwcname': new_vwc.metadata.name,
                   'namespace': ns,
                   'cmname': cm_name,
                   'hookname': hook.name,
                   'svcnamespace': hook.client_config.service.namespace,
                   'svcname': hook.client_config.service.name})
```

(Python also supports multiline strings (preserving newlines and spacing) like golang does with backticks or shell does with the `cat <<EOF` thing via triple quotes (single `'''` or double `"""`).)

Would either/both of those things (keying interpolations, cutting over to printf-style) assuage your discomfort on the "long strings" issue?

---------

> ```
> def get_validating_webhook_configuration_objects_with_annotation(client,
>                                                                  annotation):
>     ret = []
>     for o in client.list_validating_webhook_configuration().items:
>         if (o.metadata.annotations is not None and
>                 annotation in o.metadata.annotations):
>             ret.append(o)
>     return ret
> ```
> 
> In this example, I find the method arguments to be clear and net neutral with the previous code. However, in the `if` statement beginning (now) on line 48 there is a line break split on `is not None and`. When reading this the mind wants to continue reading to the right to the end of line, especially since the previous line extends 16 columns further, so the newline appears out of place (to me). As it happens, If the `if` statement were continued on one line it would only extend to 90 columns. Here, I think it isn't worth conserving 10 columns.

Yeah, I agree that splitting conditionals just sucks. (It sucks in pretty much any language, but more so in python because you can't use braces for the contents of the loop/condition even if you want to.) Again, though, you're going to have to contend with this at some point for complex conditions no matter what line length limit you choose. It's definitely not worth it to throw the whole rule away for the sake of one or two examples. And if we make the limit 90, there'll soon be a situation where we have one that's 91, etc.

For examples like this, that are just a leetle bit over the limit, there are several options, none of them beautiful.

1. Extract pieces of the condition into temp variables. While this can make the actual condition arguably more readable, I don't love the extra lines of code or throwaway variables.
```
         not_none = o.metadata.annotations is not None
         has_annotation = annotation in o.metadata.annotations
         if not_none and has_annotation:
             ret.append(o)
```
2. Make a helper method. Certainly overkill for this case -- but totally suitable for [this recent example](https://github.com/openshift/managed-cluster-validating-webhooks/pull/40/files#diff-41d387da2826fe5cb8c7c2dbf8d03061R60-R73).
```
         def has_annotation(obj, annotation):
             if obj.metadata.annotations is None:
                 return False
             return annotation in o.metadata.annotations

         if has_annotation(o, annotation):
             ret.append(o)
```
3. Tell the linter to look the other way. Python's [noqa](http://flake8.pycqa.org/en/3.2.0/user/ignoring-errors.html) is equivalent to golang's `nolint` (which I had occasion to use [recently](https://github.com/openshift-online/ocm-cli/pull/68/files#diff-091ace02b49a46f55f2586cfec89a160R40)).
```
    if o.metadata.annotations is not None and annotation in o.metadata.annotations:  # noqa: E501
      ret.append(o)
```
I've been saving this one for last because I definitely want it to be used _very_ sparingly, when it's truly the least evil way to get around the rule. In this specific case, I could probably be talked into it, though it's a toss-up between that and option 1.

---------

> Using a tabstop of 8 I feel contributes to the issue because so much of the screen width is sacrificed to white space (a Python pet peeve of mine).

A tabstop of 8 _would_ be egregious. The standard is 4. (Personally, I would enjoy 2 (which appears to be your preference) but 4 seems like a good compromise. Particularly for people who prefer vim, which doesn't (to my knowledge) add those helpful vertical lines for long blocks.) And the canonical argument for running out of horizontal space is that you're nesting too deeply and should extract some logic into helpers. (That also plays into capping cyclomatic complexity, which is a whole separate religious war.)

> A separate issue: Do we have a `.vimrc` vscode `settings.json` file (and etc) we can give to folks to help the developer-side authorship?

That would be a great idea. Something to drop in the ops-sop repo, maybe? PyCharm has most pep8-isms configured out of the box, since it was written for python from the ground up. I'm new to vscode and haven't yet figured out where those settings lie.

---------

So -- can we move forward with this if I propose an update (here or in a followon PR) to tweak the more offensive bits as suggested above?

Really appreciate the discussion here @lisa.

---------

[1] I had this exact argument about 15 years ago, from the other side. It was about C, and we didn't have a linter, so the guidelines under discussion were going to be "socially enforced". There was a guy in the room whose argument was based on the fact that he would _print out code_ to review it. Like on a printer. I told him to print landscape :P  Another guy almost hit me.
[2] That said, when it's enforced, I can fit fit four windows horizontally in my IDE, which can be super useful. I did it just the other day, in fact, when I was working on PR #51: I was following the example of test_regular_user_validation.py, so I was looking at that file and the source it was testing, as well as the file I was composing and the source _it_ was testing.
[3] This actually came up in openstack/nova quite recently: https://review.opendev.org/#/c/706311/ As you can see, it was ultimately NACKed, though mainly for reasons that don't apply here (backports).
[4] I just realized "line length limit" is probably why gofmt's rule is called 'lll' :)

2uasimojo:
@lisa, I made some of those tweaks via PR #52.

lisa:
@2uasimojo I apologize for the long delay in getting back to you, I hope you're aware of the internal things going on that are taking my time? 😉 One thing the delay has afforded me is a lot of time to think about the issues relating to linting in repositories for which I am (partly) responsible. In that time I think my position has settled.

The goal of linting should be to ensure syntactically correct content. I believe that we should focus on the human element, with code that is easy to read and easy to write. I don't want to impose so many restrictions that we spend the majority of our time fighting a linter instead of making productive code contributions. If we can use the general technologies at hand to allow developers to display the code locally how he or she wishes (think: cascading style sheets) while retaining a canonical representation here in git, I would prefer that. With that in mind,

* Column limits: Delegate to developers' editors how (if) they would like to wrap for clarity. TextMate and others offer a un/wrapping feature. With this option we aren't forced to choose an arbitrary limit: 76, 80, 90, 100, 120, whatever; we wouldn't need to append eyesore *linter directives* in our *code*. We retain the ability with peer review (the other meaning of "PR") to reign in egregious uses of this delegation. (ex., "your 4000 column line is a bit too long do you think?")
* String tokenization: I personally like the `{tokenname}` approach and think it's an improvement over the `printf(3)` style. Perhaps this will combine nicely with whatever column linting.

What do we need to do in order to get to a place of resolution on this topic?

2uasimojo:
> The goal of linting should be to ensure syntactically correct content.

That's one goal, but another is code style best practices. (You had me doubting myself, so I looked it up to make sure. Every definition "out there" includes style.)

> I don't want to impose so many restrictions that we spend the majority of our time fighting a linter instead of making productive code contributions.

Having spent the past five years using python daily in one of the largest open source projects in the world, I can tell you that the time and effort saved by enforcing linting *far* exceeds that spent fighting the linter. Once it is activated and enforced, it is very easy to get used to and consumes little bandwidth to maintain. Again, this is no different than enforcing `gofmt` (which also has a line length limit) for our go code. Do you feel like that's more trouble than it's worth?

> If we can use the general technologies at hand to allow developers to display the code locally how he or she wishes (think: cascading style sheets)

Sure, to the extent that you can choose your own font, font size, syntax highlighting scheme, etc. But we can't just have a free-for-all. Especially when my choice of local display results in me formatting in a way that your choice of local display doesn't play well with, making it harder for you to read.

> while retaining a canonical representation here in git, I would prefer that.

This actually makes the point perfectly. The code window on github is a set width (looks like about 128c) and doesn't support wrapping as far as I can see. Scrolling right is difficult, especially when the bottom of the file (and thus the horizontal scrollbar) is off the screen. Quickly catting a file from a terminal window without having to resize it, also nice. Likewise local git diff. Part of the purpose of formatting rules is to find the common denominator that may be a tiny bit uncomfortable for some people some of the time, but is almost never really uncomfortable for anyone.

> We retain the ability with peer review (the other meaning of "PR") to reign in egregious uses of this delegation. (ex., "your 4000 column line is a bit too long do you think?")

That's a nice idea, but then we wind up having variants of this discussion on every PR where style preferences clash. That *does* cost more time and energy.

> * String tokenization: I personally like the `{tokenname}` approach and think it's an improvement over the `printf(3)` style. Perhaps this will combine nicely with whatever column linting.

This is a separate area. I mentioned it because you said you had trouble tracking tokens, and thought it could help make some of the linter-enforced formatting choices more bearable. But the choice of string interpolation mechanism isn't dictated by pep8 -- we would have to enforce our own "hacking rules".

> What do we need to do in order to get to a place of resolution on this topic?

Pull in some other people?

Is line length really the only major sticking point? Can you live with the rest?

ehashman:
Just sticking my nose in here with some brief thoughts... there's a talk that might be of interest for you folks: "Beyond PEP8" https://www.youtube.com/watch?v=wf-BqAjZb8M

One my frustrations with PEP8 is that it's kind of un-Pythonic in that there's very much more than one way to do it. If you want to remove human judgment completely ala `gofmt`, the latest hotness is https://github.com/psf/black

2uasimojo:
Thanks for the input @ehashman. The issue we're having with this PR is that @lisa considers pep8 *too* strict. I wouldn't be opposed to black, but it would be even further in that direction.

black would take all the guesswork out, and can be set up to edit files in place automatically (like `gofmt` can -- though the couple of projects I've seen here don't do that).

That said, black's rules can also be customized, which would bring us right back to the same discussion we're having.

2uasimojo:
See #58 for an alternative to this, using [black](https://github.com/psf/black).

2uasimojo:
Superseded by https://github.com/openshift/managed-cluster-validating-webhooks/pull/60

or if we decide to go with black, #58.

PR #52, opened 2020-04-10
Make short lines prettier

2uasimojo:
Fix up a couple of aesthetic issues brought about by enforcing 80c
lines:
- Eliminate multiline conditions with some simple refactoring.
- Don't split multiline strings really close to the end.
- Use keys in a couple of the longer interpolated strings to make it
  easier to track which vars go where.

### DIFF
--- a/build/generate_syncset.py
+++ b/build/generate_syncset.py
@@ -1,30 +1,36 @@
 #!/usr/bin/env python
 
-import oyaml as yaml
-
 import os
-import sys
 import argparse
 import copy
 
+import oyaml as yaml
+
+
+TPL = "00-osd-managed-cluster-validating-webhooks.selectorsyncset.yaml.tmpl"
+
+
 def get_yaml_all(filename):
-    with open(filename,'r') as input_file:
+    with open(filename, 'r') as input_file:
         return list(yaml.load_all(input_file))
 
+
 def get_yaml(filename):
-    with open(filename,'r') as input_file:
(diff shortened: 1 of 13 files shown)


2uasimojo:
I still haven't figured out how to stack new commits onto an existing PR without squashing. Please review only commit 2d5c2f9a872d22edb42f4b9697ec969a259fce14 in this PR; commit def4dd47abc44d9c84a542ac874747eb693c81b7 should be reviewed in #47.

build/generate_syncset.py
```diff
@@ -1,31 +1,37 @@
...
-
 import os
-import sys
 import argparse
 import copy
 
+import oyaml as yaml
+
+
+TPL = "00-osd-managed-cluster-validating-webhooks.selectorsyncset.yaml.tmpl"
```
jewzaam, line 10: TPL sounds like a TLA, could we not shorten unless it's a very common shortening?  Recommend using `TEMPLATE` at a minimum.

2uasimojo:
> We (@jharrington22, @mwoodson, and me) discussed and are going to be looking at 120 char limit. Much of this PR is assuming a 80 char line limit and will not be necessary with 120 chars limit. Please sync w/ James on this.

Done via https://github.com/openshift/managed-cluster-validating-webhooks/pull/60 (it was easier to start over).

2uasimojo:
Superseded by #60

PR #54, opened 2020-04-17
Add validating webhook for user and identity

wanghaoran1988:
https://issues.redhat.com/browse/OSD-3271

### DIFF
--- a/src/webhook/identity_validation.py
+++ b/src/webhook/identity_validation.py
@@ -1,0 +1,80 @@
+from flask import request, Blueprint
+import sys, traceback
+import os
+from prometheus_client import Counter
+from webhook.request_helper import validate, responses
+
+bp = Blueprint("identity-webhook", __name__)
+
+# define what we track, declare Counter, how many times this route is accessed
+TOTAL_IDENTITY = Counter('webhook_identity_validation_total', 'The total number of identity validation requests')
+DENIED_IDENTITY = Counter('webhook_identity_validation_denied', 'The total number of identity validation requests denied')
+
+identity_prefix = os.getenv("IDENTITY_VALIDATION_PREFIX", "OpenShift_SRE")
+identity_prefix_list = identity_prefix.split(",")
+
+admin_group = os.getenv("GROUP_VALIDATION_ADMIN_GROUP", "osd-sre-admins,osd-sre-cluster-admins")
+admin_groups = admin_group.split(",")
+
+
+@bp.route('/identity-validation', methods=['POST'])
+def handle_request():
+  # inc total identity counter
+  TOTAL_IDENTITY.inc()
+  debug = os.getenv("DEBUG_IDENTITY_VALIDATION", "False")
+  debug = (debug == "True")
+
+  if debug:
+    print("REQUEST BODY => {}".format(request.json))
+
+  valid = True
+  try:
+    valid = validate.validate_request_structure(request.json)
+  except:
+    valid = False
+
+  if not valid:
+    # inc denied identity counter
+    DENIED_IDENTITY.inc()
+    return responses.response_invalid()
+
+  try:
+    body_dict = request.json['request']
+    # If trying to delete a user, must get user name from oldObject instead of object
+    if body_dict['object'] is None:
+      identity_name = body_dict['oldObject']['metadata']['name']
+    else:
+      identity_name = body_dict['object']['metadata']['name']
+    userinfo = body_dict['userInfo']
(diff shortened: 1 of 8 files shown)


src/webhook/identity_validation.py
```diff
@@ -0,0 +1,80 @@
...
+    userinfo = body_dict['userInfo']
+    if userinfo['username'] in ("kube:admin", "system:admin", "system:serviceaccount:openshift-authentication:oauth-openshift"):
+      # kube/system admin can do anything
+      if debug:
+        print("Performing action: {} on identity {} by {}".format(body_dict['operation'], identity_name, userinfo['username']))
+      return responses.response_allow(req=body_dict)
+ 
+    sre_user = False
+    for id_prefix in identity_prefix_list:
+      if identity_name.startswith(id_prefix):
```
jewzaam, line 57: We don't create any identity other than `OpenShift_SRE`, do we need a prefix check or just equals?

src/webhook/user_validation.py
```diff
@@ -0,0 +1,72 @@
...
+from prometheus_client import Counter
+from webhook.request_helper import validate, responses
+
+bp = Blueprint("user-webhook", __name__)
+
+# define what we track, declare Counter, how many times this route is accessed
+TOTAL_USER = Counter('webhook_user_validation_total', 'The total number of user validation requests')
+DENIED_USER = Counter('webhook_user_validation_denied', 'The total number of user validation requests denied')
+
+user_suffix = "@redhat.com"
```
jewzaam, line 13: Recommend renaming `user_suffix` to something more descriptive like `protected_user_suffix`

src/webhook/identity_validation.py
wanghaoran1988: @jew I thought we may add others for other teams like RHMI and CEE, if we don't, I will remove the list. Here I am using identity name to check the prefix, it has `OpenShift_SRE:Z3NsZWVtYW4` format, I think I should use `providerName` name, so a equals makes more sense.

wanghaoran1988:
@jewzaam Unit test added, PTAL

src/webhook/test_identity_validation.py
```diff
@@ -0,0 +1,81 @@
...
+                    "uid": "testuser",
+                    "operation": "UPDATE",
+                    "userInfo": {
+                        "groups": groups,
+                        "username": username,
+                    },
+                    "object": {
+                        "kind": "Identity",
+                        "metadata": {
+                            "name": "identityName"
```
jewzaam, line 25:
```suggestion
                            "name": identityName
```

src/webhook/test_identity_validation.py
```diff
@@ -0,0 +1,81 @@
...
+        self.assertTrue(response['response']['allowed'])
+
+    # kube:admin can update redhat user identity
+    def test_kube_admin(self):
+        response = self.runtests(
+            "kube:admin", "", "OpenShift_SRE:test", "OpenShift_SRE")
+        self.assertTrue(response['response']['allowed'])
+
+    # users in sre admin groups can update redhat user identity
+    def test_sre_groups(self):
```
jewzaam, line 66: Recommend add a test to verify SRE can edit a non-SRE identity.

src/webhook/test_identity_validation.py
```diff
@@ -0,0 +1,81 @@
...
+]
+
+
+def create_request(username, groups, identityName, providerName):
+    class FakeRequest(object):
+        def __init__(self):
+            self.json = {
+                "request": {
+                    "uid": "testuser",
+                    "operation": "UPDATE",
```
jewzaam, line 17: Should we write test to iterate over the operations we care about and just make sure we don't have regressions?  Could verify CREATE, UPDATE, and DELETE.

jewzaam:
@wanghaoran1988 same comments for test_user_validation.py as I put on test_identity_validation.py

src/webhook/test_identity_validation.py
wanghaoran1988: No, we can only do it in e2e tests, this is only unit test, the validation func doesn't check whether it's "CREATE" or "UPDATE", the verbs are configured in `ValidatingWebhookConfiguration`.

src/webhook/test_identity_validation.py
wanghaoran1988: done

src/webhook/test_identity_validation.py
wanghaoran1988: good catch

wanghaoran1988:
@jewzaam Comments addressed, could you take another look?

PR #58, opened 2020-04-28
Lint with black

2uasimojo:
This switches from `flake8` to [black](https://github.com/psf/black)
linting, including adding a `black` make target you can run to get your
code compliant.

This is in the same spirit as how other projects lint go code:
- Like `gofmt`, `black` is highly opinionated.
- Like `gofmt`, `black` has the ability to reformat and rewrite code in
place. But
- Like our projects that lint-check go with `gofmt`, we run this in
advisory mode and simply fail the build if there are deviations from
what `black` wants, whereupon
- Like our projects that lint-check go with `gofmt`, it is the
committer's responsibility to run `black` in overwrite mode (which can
be done via `make black`) and commit the changes in order to pass the
build.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,14 +34,11 @@ test-container:
 
 .PHONY: lint
 lint: test-container
-	# E111 indentation is not a multiple of four
-	# E121 continuation line under-indented for hanging indent
-	# E401 multiple imports on one line
-	# E402 module level import not at top of file
-	# E501 line too long (N > 79 characters)
-	# E722 do not use bare 'except'
-	# W293 blank line contains whitespace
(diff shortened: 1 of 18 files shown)


c-e-brumm:
I see no issues so long as black is the new formatter we've decided on.

jewzaam:
/hold for decisions on python linting standard
cc @jharrington22

2uasimojo:
By default, black uses 88c lines (but not strictly). Should I put in a config override to 120?

2uasimojo:
It has been decreed that we shall go with flake8 with an exception for 120c lines. See #60.

PR #59, opened 2020-04-28
Move flake8 options to config

2uasimojo:
This is a straight refactor of the `lint` target to move the flake8
options from the command line into a config file. This makes it easier
to document and change them, and will make it easier to centralize this
configuration across multiple projects once we've decided on a common
standard.


lisa:
Obsolete with the merging of #78 and so closing this.

PR #60, opened 2020-04-28
Enforce all of flake8 with 120c lines

2uasimojo:
This is an alternative to https://github.com/openshift/managed-cluster-validating-webhooks/pull/52 that un-ignores all of the flake8 rules but overrides the default line width limit to 120c. Source files are reformatted accordingly.


lisa:
Obsolete with the merging of #78 and so closing this.

PR #63, opened 2020-05-04
Add unit tests for group validation webhook OSD-3361

csheremeta:
Name                                          Stmts   Miss  Cover   Missing
---------------------------------------------------------------------------
src/webhook/group_validation.py                  55     17    69%   22-40, 55, 59, 68

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ A Flask app designed to act as a webhook admission controller for OpenShift.
 
 Configuration for this webhook is provided by environment variables:
 
-* `GROUP_VALIDATION_PREFIX` - Group prefix to apply the webhook, such as `osd-` to apply to `CREATE`, `UPDATE`, `DELETE` operations on groups starting with `osd-`. (default: `osd-sre-`)
+* `GROUP_VALIDATION_PROTECTED_GROUP_REGEX` - Regular expression for protected group names, such as `osd-` to apply to `CREATE`, `UPDATE`, `DELETE` operations on groups starting with `osd-`. (default: `(^osd-sre.*|^dedicated-admins$|^cluster-admins$|^layered-cs-sre-admins$)`)
 * `GROUP_VALIDATION_ADMIN_GROUP` - Admin groups, which the requestor must be a member in order to have access granted. This is comma-separated. (default: `osd-sre-admins,osd-sre-cluster-admins`)
 * `DEBUG_GROUP_VALIDATION` - Debug the webhook? Set to `True` to enable, all other values (including absent) disable. (default: False)
 
--- a/src/webhook/group_validation.py
+++ b/src/webhook/group_validation.py
@@ -1,6 +1,7 @@
 from flask import request, Blueprint
 import sys, traceback
 import os
+import re
 from prometheus_client import Counter
 from webhook.request_helper import validate, responses
 
@@ -10,10 +11,12 @@ bp = Blueprint("group-webhook", __name__)
 TOTAL_GROUP = Counter('webhook_group_validation_total', 'The total number of group validation requests')
 DENIED_GROUP = Counter('webhook_group_validation_denied', 'The total number of group validation requests denied')
 
-group_prefix = os.getenv("GROUP_VALIDATION_PREFIX", "osd-sre-")
 admin_group = os.getenv("GROUP_VALIDATION_ADMIN_GROUP", "osd-sre-admins,osd-sre-cluster-admins")
 
 admin_groups = admin_group.split(",")
+
+# groups that cannot be edited by non-admins
+protected_group_regex = os.getenv("GROUP_VALIDATION_PROTECTED_GROUP_REGEX", "(^osd-sre.*|^dedicated-admins$|^cluster-admins$|^layered-cs-sre-admins$)")
 
 
 @bp.route('/group-validation', methods=['POST'])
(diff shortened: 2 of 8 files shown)


csheremeta:
now:
src/webhook/group_validation.py                  64     17    73%   25-43, 69, 75, 87
(thanks @jewzaam )

jewzaam:
Since I applied a pretty big change on top of this in https://github.com/csheremeta/managed-cluster-validating-webhooks/pull/1 I abstain from review.

Note the layered-cs-sre-admins stuff partially addresses https://issues.redhat.com/browse/OSD-2944 and I'll wait to do any more tweaks until this PR lands.

src/webhook/group_validation.py
```diff
@@ -37,34 +40,54 @@ def handle_request():
...
+    userinfo = {}
+    if 'userInfo' in body_dict:
+      userinfo = body_dict['userInfo']
+    user_groups = []
+    if 'groups' in userinfo:
+      user_groups = userinfo['groups']
+
+    user_name = None
+    if 'username' in userinfo:
+      user_name = userinfo['username']
```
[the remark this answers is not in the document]
jewzaam: Yea, should be replaced.  I did this then updated validation a little but didn't want to go too far.  We could probably validate that a username exists and just trust if we get this far it's ok.

src/webhook/namespace_validation.py
```diff
@@ -40,7 +40,12 @@ def handle_request():
 def get_response(req, debug=False):
   try:
     body_dict = req.json['request']
-    requester_group_memberships = body_dict['userInfo']['groups']
+    requester_group_memberships = []
+    if 'userInfo' in body_dict:
+      if 'groups' in body_dict['userInfo']:
+        requester_group_memberships = body_dict['userInfo']['groups']
+    else:
+      return responses.response_invalid()
```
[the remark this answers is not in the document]
jewzaam: Yes, it's necessary and was causing tests to fail.  There was some weirdness with tests using the same names for global scoped vars that was masking this until I moved the group validation test global vars into a class (and then namespace validation test global vars).

README.md
```diff
@@ -8,7 +8,7 @@ A Flask app designed to act as a webhook admission controller for OpenShift.
 
 Configuration for this webhook is provided by environment variables:
 
-* `GROUP_VALIDATION_PREFIX` - Group prefix to apply the webhook, such as `osd-` to apply to `CREATE`, `UPDATE`, `DELETE` operations on groups starting with `osd-`. (default: `osd-sre-`)
+* `GROUP_VALIDATION_PROTECTED_GROUP_REGEX` - Regular expression for protected group names, such as `osd-` to apply to `CREATE`, `UPDATE`, `DELETE` operations on groups starting with `osd-`. (default: `(^osd-sre.*|^dedicated-admins$|^cluster-admins$|^layered-cs-sre-admins$)`)
```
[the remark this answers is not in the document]
jewzaam:
```suggestion
* `GROUP_VALIDATION_PROTECTED_GROUP_REGEX` - Regular expression for groups that cannot be edited by non-SRE users, such as `osd-` to apply to `CREATE`, `UPDATE`, `DELETE` operations on groups starting with `osd-`. (default: `(^osd-sre.*|^dedicated-admins$|^cluster-admins$|^layered-cs-sre-admins$)`)
```

jewzaam:
So, I could merge one a suggestion and did but it's not my PR.. sorry!  Didn't know I could do that.

csheremeta:
/hold cancel
@jewzaam @lisa I added more test cases to fully cover the logic introduced by Naveen's commits, PTAL

csheremeta:
@lisa I don't see value in adding a create_create_request function. the only reason I separated out the create_request from create_delete_request is because the delete requests look at the oldObject metadata, instead of the object metadata used by create/update/etc requests. the string represented in either function's "operation" tag could be "bobsyouruncle" for all it matters - this difference was not the point of the exercise

jewzaam:
@csheremeta at this time the operation doesn't matter to the code so I'm in agreement.  The only benefit I can see to adding a "create" will be future proof for the unit tests.  My recommendation is to not hold up this PR but to create a new jira card for adding "create" vs "update" tests for all the unit tests in this repo.  I know we don't have it for everything, so rather it's covered for everything.

jewzaam:
Looks reasonable to me, will let someone else +1 it though given my PR into this, so if @lisa could review it would be great.

/approve

src/webhook/group_validation.py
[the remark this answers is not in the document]
lisa: `userInfo` [is a part of the required spec](https://github.com/kubernetes/api/blob/master/admission/v1/types.go#L94-L95) for an `AdmissionReview`, but within the `UserInfo` struct, everything (including the `username`) [is optional](https://github.com/kubernetes/api/blob/master/authentication/v1/types.go#L100-L117). So checking for the username and groups is probably where we want to be.

src/webhook/test_namespace_validation.py
```diff
@@ -4,52 +4,6 @@
 from webhook import namespace_validation
 
 
-PRIVILEGED_NAMESPACES = (
```
lisa, line 7: Why are the namespace tests changing in a PR about group validation tests? Will these changes introduce merge conflicts with https://github.com/openshift/managed-cluster-validating-webhooks/pull/66 (cc @jewzaam)?

src/webhook/test_namespace_validation.py
```diff
@@ -67,6 +21,51 @@ class FakeRequest(object):
 
 
 class TestNamespaceValidation(unittest.TestCase):
+    PRIVILEGED_NAMESPACES = (
+        "kube-admin",
+        "kube-foo",
+        "kube-",
```
lisa, line 27:
```suggestion
```

To be in line with changes in https://github.com/openshift/managed-cluster-validating-webhooks/pull/66, I'll make the same suggested changes here.

`kube-` is not a valid Namespace name, and so it doesn't make sense to test it. The Kubernetes API would deny the creation of such a Namespace.

src/webhook/test_namespace_validation.py
```diff
@@ -67,6 +21,51 @@ class FakeRequest(object):
...
+        "openshift",
+        "openshifter",
+        "openshift-foo",
+        "ops-health-monitoring",
+        "management-infra",
+        "default",
+        "logging",
+        "sre-app-check",
+        "redhat-user",
+        "redhat-",
```
lisa, line 37: To be in line with changes in https://github.com/openshift/managed-cluster-validating-webhooks/pull/66, I'll make the same suggested changes here.

`redhat-` is not a valid Namespace name, and so it doesn't make sense to test it. The Kubernetes API would deny the creation of such a Namespace.
```suggestion
```

src/webhook/test_namespace_validation.py
csheremeta: done

src/webhook/test_namespace_validation.py
csheremeta: done

PR #65, opened 2020-05-05
:bug: When pwd involves a symlink, test, build-sss, lint fail

lisa:
When `pwd` includes a symlink, the `test`, `build-sss` and `lint` make targets fail because even though the correct volume is mounted (with `-v` and using `pwd -P`), the "cd" inside the container does not respect it, and so will try to `cd` to the wrong place.


lisa:
/kind bug

oops

PR #67, opened 2020-05-07
CSSRE-998: Added support for protecting exclusive groups.

pepedocs:
Jira: https://issues.redhat.com/browse/CSSRE-988

**Problem**
Some "exclusive" groups (see jira) are not protected by the group validation. For example, currently, there is no way to enforce that a group `my-group-a` can only be operated on (create, update, delete) by a user that is a member of a group with prefix `my-group`.

> Note that the DONE requirements are in Jira, please have a look for the specific group examples.

**Solution**
- Introduced an env var `GROUP_VALIDATION_EXCLUSIVE_GROUP_PREFIXES` that will be used to specify a list of "exclusive" groups.
- Added additional check to disallow user from performing operations to exclusive groups if he/she isn't not a member of a group that has the prefix inlcuded in the env var.
- Also enhanced the readability of some code (e.g. debug)

**Test**
- A unit test is added in `test_group_validation.test_exclusive_group` to ensure the new logic
- All tests passed, if needed I can attach the logs
- The tests can be simulated at unit-test-level, therefore I have not tested this in an OSD cluster as it will probably take so much effort just to test a simple logic.

Notes
- Creating a new webhook would also involve so much unnecessary effort
- The env var will need to be updated for the pods/deployment to be able to use this feature
- Note that this specific check is on top of the existing logic. This will also disallow even admins to operate on exclusive groups (as literally described in Jira)

### DIFF
--- a/src/webhook/group_validation.py
+++ b/src/webhook/group_validation.py
@@ -7,6 +7,8 @@ from webhook.request_helper import validate, responses
 
 bp = Blueprint("group-webhook", __name__)
 
+DEBUG = os.getenv("DEBUG_GROUP_VALIDATION", "False")
+
 # define what we track, declare Counter, how many times this route is accessed
 TOTAL_GROUP = Counter('webhook_group_validation_total', 'The total number of group validation requests')
 DENIED_GROUP = Counter('webhook_group_validation_denied', 'The total number of group validation requests denied')
@@ -18,16 +20,24 @@ admin_groups = admin_group.split(",")
 # groups that cannot be edited by non-admins
 protected_group_regex = os.getenv("GROUP_VALIDATION_PROTECTED_GROUP_REGEX", "(^osd-sre.*|^dedicated-admins$|^cluster-admins$|^layered-cs-sre-admins$)")
 
+exclusive_group_prefixes = os.getenv("GROUP_VALIDATION_EXCLUSIVE_GROUP_PREFIXES", "").split(",")
+
+
+def log_debug(msg):
+  if DEBUG:
+    print(msg)
+
+
+def log(msg):
+  print(msg)
+
 
 @bp.route('/group-validation', methods=['POST'])
 def handle_request():
   # inc total group counter
   TOTAL_GROUP.inc()
-  debug = os.getenv("DEBUG_GROUP_VALIDATION", "False")
-  debug = (debug == "True")
 
-  if debug:
-    print("REQUEST BODY => {}".format(request.json))
+  log_debug("REQUEST BODY => {}".format(request.json))
 
   valid = True
   try:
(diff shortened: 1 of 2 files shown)


pepedocs:
Ping @alanmoran

alanmoran:
Seems like we had two PR's for the same thing https://github.com/openshift/managed-cluster-validating-webhooks/pull/67. @pepedocs Can you review and determine if we need to make any further changes? 

CC @jewzaam FYI

pepedocs:
@alanmoran It looks like #63 was simply adding a check for "protected" groups = (^osd-sre.*|^dedicated-admins$|^cluster-admins$|^layered-cs-sre-admins$) before performing the validation, plus some unit tests to test group validation. With #63, it still has the following behavior

- osd-sre* can operate on layered-* groups

This contradicts (at least my understanding) with the requirement in the ticket that only users in layered-* can operate on layered-* groups, I know osd-* is much more privileged but I took the requirement literally.

Otherwise, if we decide close this PR because #63 is enough I think we still need to add layered-* in the admins env var in the pod's deployment.

alanmoran:
> @alanmoran It looks like #63 was simply adding a check for "protected" groups = (^osd-sre.*|^dedicated-admins$|^cluster-admins$|^layered-cs-sre-admins$) before performing the validation, plus some unit tests to test group validation. With #63, it still has the following behavior
> 
> * osd-sre* can operate on layered-* groups
> 
> This contradicts (at least my understanding) with the requirement in the ticket that only users in layered-* can operate on layered-* groups, I know osd-* is much more privileged but I took the requirement literally.
> 
> Otherwise, if we decide close this PR because #63 is enough I think we still need to add layered-* in the admins env var in the pod's deployment.

@pepedocs being honest - SRE P own the base platform. I don't have an overly large concern about this. They have the same if not, more permissions than CS SRE. Happy to update the requirement if you like?

pepedocs:
> > @alanmoran It looks like #63 was simply adding a check for "protected" groups = (^osd-sre.*|^dedicated-admins$|^cluster-admins$|^layered-cs-sre-admins$) before performing the validation, plus some unit tests to test group validation. With #63, it still has the following behavior
> > 
> > * osd-sre* can operate on layered-* groups
> > 
> > This contradicts (at least my understanding) with the requirement in the ticket that only users in layered-* can operate on layered-* groups, I know osd-* is much more privileged but I took the requirement literally.
> > Otherwise, if we decide close this PR because #63 is enough I think we still need to add layered-* in the admins env var in the pod's deployment.
> 
> @pepedocs being honest - SRE P own the base platform. I don't have an overly large concern about this. They have the same if not, more permissions than CS SRE. Happy to update the requirement if you like?

Let's close this then!

PR #68, opened 2020-05-07
Fix 2 linting errors: W291 trailing whitespace

jewzaam:


jewzaam:
I was surprised to find `make` failed, there were two errors in linting.  Means builds are broken right now..

lisa:
Non-functional whitespace should not cause builds to fail.

/lgtm

PR #69, opened 2020-05-07
Remove blank line that's causing build failures

csheremeta:


lisa:
/lgtm

The previous one looked good to me too, so 🤷‍♀️

PR #70, opened 2020-05-08
This fix the request validation issue

wanghaoran1988:
The "userInfo" is not part of the top level key, this will fail all the validations.


csheremeta:
manually merged as breakfix. @jewzaam please review this in the morning, as you may want to resubmit the code

PR #62, opened 2020-04-30
Change logic for namespace validation according to OSD-3511

csheremeta:
/hold for testing
https://issues.redhat.com/browse/OSD-3511

PR #72, opened 2020-05-08
Merge OSD-3511 changes

jewzaam:
https://issues.redhat.com/browse/OSD-3511
blocks: https://issues.redhat.com/browse/INTLY-7632

Merges https://github.com/openshift/managed-cluster-validating-webhooks/pull/62
Fixes some validation holes
Adds some missing increment of deny counters (metrics)
A few lint fixes
Add 'lint' and 'test' make targets to default 'all' make target

### DIFF
--- a/src/webhook/namespace_validation.py
+++ b/src/webhook/namespace_validation.py
@@ -40,25 +40,45 @@ def handle_request():
 def get_response(req, debug=False):
   try:
     body_dict = req.json['request']
-    requester_group_memberships = []
-    if 'userInfo' in body_dict:
-      if 'groups' in body_dict['userInfo']:
-        requester_group_memberships = body_dict['userInfo']['groups']
-    else:
-      return responses.response_invalid()
-    if "dedicated-admins" in requester_group_memberships:
-      requested_ns = body_dict['namespace']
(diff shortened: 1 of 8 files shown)


jewzaam:
tide blocking though it says good to merge
could be related to https://github.com/openshift/release/pull/8947
need this to land, so merging

src/webhook/namespace_validation.py
```diff
@@ -40,25 +40,45 @@ def handle_request():
...
+    cluster_admin_users = ["kube:admin", "system:admin"]
+
+    # check to see if requester is a serviceAccount in a privileged NS
+    # if so, SA can edit any namespace
+    # re.match will return a match object if the namespace matches the regex,
+    # or None if the string doesn't match the regex
+    privileged_serviceaccount_re = '^system:serviceaccounts:(kube.*|openshift.*|default|redhat.*)'
+    for group in requester_group_memberships:
+      if re.match(privileged_serviceaccount_re, group) is not None:
         return responses.response_allow(req=body_dict)
```
lisa, line 56: @jewzaam This is well past merge, but it looks like this is matching the service account name to user group names? Notably, this section of code does not have tests, which could help with the previous question

csheremeta: I talked to Lisa about this a little earlier today, since this is my code. She's right that there's no unit tests for this bit. I'll create a PR with unit tests for it.

PR #74, opened 2020-05-14
freeze-window-validation-webhook-tmpl - OSD-3515

himanshudogra:

PR #77, opened 2020-05-24
OSD-3514 - Add helper script to fetch UpgradeConfig resource

ravitri:
Add helper script to fetch UpgradeConfig resource for PreferredUpgradeWindow and FreezeWindow Validating webhooks.

### DIFF
--- a/src/webhook/upgrade_operator/fetch_upgradeconfig.py
+++ b/src/webhook/upgrade_operator/fetch_upgradeconfig.py
@@ -1,12 +1,16 @@
 import os
 from kubernetes import client, config
 from kubernetes.client.rest import ApiException
+from kubernetes.config.config_exception import ConfigException
 
 # Use KUBECONFIG from pod else consume from local ~/.kube/config
-if 'KUBERNETES_PORT' in os.environ:
-    config.load_incluster_config()
-else:
-    config.load_kube_config()
+try:
+    if 'KUBERNETES_PORT' in os.environ:
+        config.load_incluster_config()
+    else:
+        config.load_kube_config()
+except ConfigException as c:
+    print("Exception in loading cluster configuration : %s\n" % c)
 
 
 def get_upgradeconfig_cr():
@@ -14,6 +18,8 @@ def get_upgradeconfig_cr():
     try:
         ugpradeconfig_resource = custom_api.list_cluster_custom_object(
             'upgrade.managed.openshift.io', 'v1alpha1', 'upgradeconfigs')
+        if ugpradeconfig_resource is None:
+            return None
         return ugpradeconfig_resource
     except ApiException as e:
         print(
--- a/src/webhook/upgrade_operator/fetch_upgradeconfig.py
+++ b/src/webhook/upgrade_operator/fetch_upgradeconfig.py
@@ -1,0 +1,20 @@
+import os
+from kubernetes import client, config
+from kubernetes.client.rest import ApiException
+
+# Use KUBECONFIG from pod else consume from local ~/.kube/config
+if 'KUBERNETES_PORT' in os.environ:
+    config.load_incluster_config()
+else:
+    config.load_kube_config()
+
+
+def get_upgradeconfig_cr():
+    custom_api = client.CustomObjectsApi()
+    try:
+        ugpradeconfig_resource = custom_api.list_cluster_custom_object(
+            'upgrade.managed.openshift.io', 'v1alpha1', 'upgradeconfigs')
+        return ugpradeconfig_resource
+    except ApiException as e:
+        print(
+            "Exception in fetching upgrade.managed.openshift.io/v1alpha1 resource: %s\n" % e)
--- a/src/webhook/upgrade_operator/fetch_upgradeconfig.py
+++ b/src/webhook/upgrade_operator/fetch_upgradeconfig.py
@@ -1,26 +1,40 @@
-import os
 from kubernetes import client, config
 from kubernetes.client.rest import ApiException
 from kubernetes.config.config_exception import ConfigException
 
-# Use KUBECONFIG from pod else consume from local ~/.kube/config
-try:
-    if 'KUBERNETES_PORT' in os.environ:
-        config.load_incluster_config()
+
+def kube_auth():
+    # Use ~/.kube/config for authentication if script run directly
+    if __name__ == "__main__":
+        try:
+            config.load_kube_config()
+            return True
+        except ConfigException as c:
+            print("ConfigException in load_kube_config() : %s\n" % c)
+            return False
+
+    # Use serviceaccount token to authenticate.
     else:
-        config.load_kube_config()
-except ConfigException as c:
-    print("Exception in loading cluster configuration : %s\n" % c)
+        try:
+            config.load_incluster_config()
+            return True
+        except ConfigException as c:
+            print("ConfigException in load_incluster_config() : %s\n" % c)
+            return False
 
 
 def get_upgradeconfig_cr():
-    custom_api = client.CustomObjectsApi()
-    try:
-        ugpradeconfig_resource = custom_api.list_cluster_custom_object(
-            'upgrade.managed.openshift.io', 'v1alpha1', 'upgradeconfigs')
-        if ugpradeconfig_resource is None:
+    if kube_auth():
+        custom_api = client.CustomObjectsApi()
+        try:
+            ugpradeconfig_resource = custom_api.list_cluster_custom_object(
+                'upgrade.managed.openshift.io', 'v1alpha1', 'upgradeconfigs')
+            return ugpradeconfig_resource
+        except ApiException as e:
+            print(
+                "Exception in fetching upgrade.managed.openshift.io/v1alpha1 resource: %s\n" % e)
             return None
-        return ugpradeconfig_resource
-    except ApiException as e:
-        print(
-            "Exception in fetching upgrade.managed.openshift.io/v1alpha1 resource: %s\n" % e)
+
+
+if __name__ == "__main__":
+    get_upgradeconfig_cr()
--- a/src/webhook/upgrade_operator/test_fetch_upgradeconfig.py
+++ b/src/webhook/upgrade_operator/test_fetch_upgradeconfig.py
@@ -1,0 +1,23 @@
+import unittest
+from webhook.upgrade_operator import fetch_upgradeconfig
+
+
+class TestFetchUpgradeConfig(unittest.TestCase):
+
+    def validate_upgraceconfig_object(self, request_json):
+        doc_keys = request_json.keys()
+
+        if 'kind' not in doc_keys:
+            return False
+
+        if request_json['items'][0]['kind'] == "UpgradeConfig":
+            return True
+        return False
+
+    def test_fetch_upgradeconfig_object(self):
+        if fetch_upgradeconfig.get_upgradeconfig_cr() is None:
+            self.skipTest(
+                "Skipping unit test for test_fetch_upgradeconfig.py. To be tested in e2etest.")
+        else:
+            self.assertTrue((self.validate_upgraceconfig_object(
+                fetch_upgradeconfig.get_upgradeconfig_cr())), "Valid UpgradeConfig resource")
--- a/src/webhook/upgrade_operator/test_fetch_upgradeconfig.py
+++ b/src/webhook/upgrade_operator/test_fetch_upgradeconfig.py
@@ -4,20 +4,7 @@ from webhook.upgrade_operator import fetch_upgradeconfig
 
 class TestFetchUpgradeConfig(unittest.TestCase):
 
-    def validate_upgraceconfig_object(self, request_json):
-        doc_keys = request_json.keys()
-
-        if 'kind' not in doc_keys:
-            return False
-
-        if request_json['items'][0]['kind'] == "UpgradeConfig":
-            return True
-        return False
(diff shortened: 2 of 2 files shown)


src/webhook/upgrade_operator/fetch_upgradeconfig.py
```diff
@@ -0,0 +1,20 @@
+import os
+from kubernetes import client, config
+from kubernetes.client.rest import ApiException
+
+# Use KUBECONFIG from pod else consume from local ~/.kube/config
+if 'KUBERNETES_PORT' in os.environ:
```
2uasimojo, line 6: It would be more pythonic (and only a *teeny* bit less efficient) to do this like:

```python
try:
    config.load_incluster_config()
except ConfigException:
    config.load_kube_config()
```

2uasimojo: Also, it’s concerning that you’re doing this at the file level, meaning that it’ll happen at import time. Are you sure that’s what you want to do?

src/webhook/upgrade_operator/fetch_upgradeconfig.py
```diff
@@ -0,0 +1,20 @@
...
+
+def get_upgradeconfig_cr():
+    custom_api = client.CustomObjectsApi()
+    try:
+        ugpradeconfig_resource = custom_api.list_cluster_custom_object(
+            'upgrade.managed.openshift.io', 'v1alpha1', 'upgradeconfigs')
+        return ugpradeconfig_resource
+    except ApiException as e:
+        print(
+            "Exception in fetching upgrade.managed.openshift.io/v1alpha1 resource: %s\n" % e)
```
2uasimojo, line 20: This will implicitly return `None`. That would be better to make explicit.

src/webhook/upgrade_operator/fetch_upgradeconfig.py
```diff
@@ -0,0 +1,20 @@
...
+from kubernetes.client.rest import ApiException
+
+# Use KUBECONFIG from pod else consume from local ~/.kube/config
+if 'KUBERNETES_PORT' in os.environ:
+    config.load_incluster_config()
+else:
+    config.load_kube_config()
+
+
+def get_upgradeconfig_cr():
```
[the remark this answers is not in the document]
ravitri: @2uasimojo - Thanks a lot for review! Yes, this script is meant to be a helper script for the Validating webhooks which are to be added as per https://issues.redhat.com/browse/OSD-3513 and https://issues.redhat.com/browse/OSD-3515. Both those Webhooks will require the UpgradeConfig resource which this script would fetch.

src/webhook/upgrade_operator/fetch_upgradeconfig.py
ravitri: Agreed. Actually, I was thinking whether to include the `config.load_kube_config()` or not as the client authentication would always happen using `config.load_incluster_config()` (from within the pod). But while testing the script this would be helpful when connected to a cluster remotely using `oc` client so included it anyway.

src/webhook/upgrade_operator/fetch_upgradeconfig.py
ravitri: Sorry, I didn't quite understand this comment. Could you elaborate? If you are referring to whether the script the method does return a valid dict object or not, it does. I am able to fetch the value in the calling script which uses this helper script.

ravitri: Ok, I think I understand your point. I will add checks to return `None` explicitly. Thanks a lot @2uasimojo !

src/webhook/upgrade_operator/fetch_upgradeconfig.py
ravitri: Have added `ConfigException` to check for kube auth

src/webhook/upgrade_operator/fetch_upgradeconfig.py
ravitri: Done

src/webhook/upgrade_operator/test_fetch_upgradeconfig.py
```diff
@@ -0,0 +1,23 @@
...
+
+    def validate_upgraceconfig_object(self, request_json):
+        doc_keys = request_json.keys()
+
+        if 'kind' not in doc_keys:
+            return False
+
+        if request_json['items'][0]['kind'] == "UpgradeConfig":
+            return True
+        return False
```
2uasimojo, line 15: So this is a bit odd, keeping in mind that I'm not sure what the JSON payload is supposed to look like...

First, just a minor thing: You don't need the `.keys()` to find out if a key is in a dict. You can just say `if 'kind' not in request_json:`. (In py3 `.keys()` just redundant, but in py2 you would actually be creating a new list of the keys.)

But mainly: You're checking that the top-level object has a `kind`. I think that's unnecessary. It's going to be a `List`, even if it has zero entries, unless you got an `ApiException`, in which case `request_json` will be `None` and attempting to access `.keys()` (or use `in` directly) will give you a `TypeError`. So I would suggest starting off with a `None` check:

```python
if request_json is None:
    return False
```

After that, it looks like you want to make sure there's at least one item, and that *its* `kind` exists and equals `UpgradeConfig`. What you've got will work as long as there's an `items` key whose value is a list with at least one element, and that element is a dict containing a `kind` key. That's probably fine for purposes of this test, since any exceptions will cause a test failure anyway (if not a pretty one). But FYI if you wanted to be really precise about it, you could do things like:

```python
itemList = request_json.get('items')
if not itemList:
    # This will trigger if request_json['items'] was None, or an empty
    # list, or (due to our `.get`) if there was no `items` key
    return False
return itemList[0].get('kind') == 'UpgradeConfig'
```

(Using `.get()` lets you access a dict key in a way that doesn't raise an exception if the key doesn't exist. Instead you get to specify a default for that case. The default default is `None`.)

src/webhook/upgrade_operator/fetch_upgradeconfig.py
```diff
@@ -1,19 +1,25 @@
 import os
 from kubernetes import client, config
 from kubernetes.client.rest import ApiException
+from kubernetes.config.config_exception import ConfigException
 
 # Use KUBECONFIG from pod else consume from local ~/.kube/config
```
2uasimojo, line 6: I'm still concerned that this is being done at the file scope. Side-effecty imports should be used very sparingly, and I feel it's not justified in this case.

2uasimojo: [Later] Actually, this is triggering your CI failure (although IMO this indicates a bug in the kubernetes client library, which should be catching that condition and raising `ConfigException` instead).

So yeah, I think you should wrap this setup in a method. If you want singleton behavior, track it in a global variable.

That would also make testing easier in general, because you could mock the client.

ravitri:
@2uasimojo - This is great feedback! I admit my Python skills need more experience and your comments are certainly helpful in me having different approaches and perspectives to same code. I will review this soon. 

However, there are some changes in the priorities for now for the upgrade operator so I might have to keep this PR on hold for a while and get back to it later on.. This will be decided/finalised tomorrow.

lisa:
Is this related to, and can it be included in #74?

ravitri:
@lisa - Yes, for now it was decided that this particular PR (https://issues.redhat.com/browse/OSD-3514) will need to be finished individually which will be referred in future by https://issues.redhat.com/browse/OSD-3513 and https://issues.redhat.com/browse/OSD-3515 (#74) webhooks. The logic to implement both of these webhooks is still not finalized but we know that both of them need this particular PR eventually. I will work on this PR soon and finish it. We can expect the webhooks logic to be finalized in next phase of upgrade operator or so. 

As far as this PR is concerned, we decided that we will have any relevant test as part of e2e test for this PR and skip the unit test for now because of python's kubernetes-client limitations. 

Let me know if any other details required.

src/webhook/upgrade_operator/fetch_upgradeconfig.py
```diff
@@ -1,19 +1,25 @@
...
-if 'KUBERNETES_PORT' in os.environ:
-    config.load_incluster_config()
-else:
-    config.load_kube_config()
+try:
+    if 'KUBERNETES_PORT' in os.environ:
```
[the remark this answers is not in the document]
ravitri: Sure, have fixed this logic to separate out the auth from within pod and outside pod. Having auth using ~/.kube/config is as such not to be part of the code since it is only meant to run within the pod but to locally run the script, have added this part for now. It's optional part.

src/webhook/upgrade_operator/fetch_upgradeconfig.py
```diff
@@ -1,19 +1,25 @@
...
+    print("Exception in loading cluster configuration : %s\n" % c)
 
 
 def get_upgradeconfig_cr():
     custom_api = client.CustomObjectsApi()
     try:
         ugpradeconfig_resource = custom_api.list_cluster_custom_object(
             'upgrade.managed.openshift.io', 'v1alpha1', 'upgradeconfigs')
+        if ugpradeconfig_resource is None:
+            return None
```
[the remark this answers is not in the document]
ravitri: Yup, agreed. I have fixed this.

src/webhook/upgrade_operator/test_fetch_upgradeconfig.py
```diff
@@ -0,0 +1,23 @@
+import unittest
+from webhook.upgrade_operator import fetch_upgradeconfig
+
+
+class TestFetchUpgradeConfig(unittest.TestCase):
+
+    def validate_upgraceconfig_object(self, request_json):
```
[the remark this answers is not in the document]
ravitri: Thanks for catching that! But I removed all unit test logic for now as it was decided to cover it in e2e tests.

src/webhook/upgrade_operator/test_fetch_upgradeconfig.py
ravitri: This part if really helpful. I will keep this in mind for future. Thanks a lot for the tip :) But this code is removed and is to be handled as part of e2e test.

src/webhook/upgrade_operator/fetch_upgradeconfig.py
ravitri: I have added changes. Can you review from Python point of view? Have tried to incorporate the changes from what I understood from your suggestions :) Let me know if it's acceptable with you.

ravitri:
@wanghaoran1988  - Could you please review and approve? Thought of covering this in APAC time tomorrow if possible.

wanghaoran1988:
I see we start to migrate from python to go https://github.com/openshift/managed-cluster-validating-webhooks/pull/78
and as we are not clear how we will go with the scheduling part, let's hold this for now?

ravitri:
@wanghaoran1988 - I wasn't aware of #78 but was looking forward to it and seems better path ahead for sure. Agree with you. If Golang webhooks get in place, I don't see the need to have this PR at all. It makes use of python's kubernetes client which has limitations regarding unit/mock tests and v1beta1 version for validatingwebhookconfiguration. Will be much better to have Golang based webhooks anyway and importing UpgradeConfig to any of the scheduling time window webhooks won't require separate helper script in Golang so am inclined to close this PR. However, will just keep it on hold for some more days.

src/webhook/upgrade_operator/fetch_upgradeconfig.py
```diff
@@ -0,0 +1,40 @@
+from kubernetes import client, config
+from kubernetes.client.rest import ApiException
+from kubernetes.config.config_exception import ConfigException
+
+
+def kube_auth():
+    # Use ~/.kube/config for authentication if script run directly
+    if __name__ == "__main__":
```
2uasimojo, line 8: I like this

src/webhook/upgrade_operator/fetch_upgradeconfig.py
```diff
@@ -0,0 +1,40 @@
...
+        try:
+            config.load_kube_config()
+            return True
+        except ConfigException as c:
+            print("ConfigException in load_kube_config() : %s\n" % c)
+            return False
+
+    # Use serviceaccount token to authenticate.
+    else:
```
2uasimojo, line 17: Note that this `else` is redundant, because the code under the `if` always returns.

Stylistically, though, it's not a terrible thing to keep it.

src/webhook/upgrade_operator/fetch_upgradeconfig.py
```diff
@@ -0,0 +1,40 @@
...
+        try:
+            config.load_incluster_config()
+            return True
+        except ConfigException as c:
+            print("ConfigException in load_incluster_config() : %s\n" % c)
+            return False
+
+
+def get_upgradeconfig_cr():
+    if kube_auth():
```
2uasimojo, line 27: I'm noticing that the result of any exception is a printed message (which would go into the container logs somewhere?) and a `None` return (which is sometimes implicit, e.g. in the case of failed auth). That may be okay depending on how this is supposed to be used by its consuming code. But it's an esoteric enough pattern that it would be nice to have it spelled out in a docstring.

ravitri:
As per https://issues.redhat.com/browse/OSD-4110, this PR might not be required at all to be implemented using python kubernetes client. Marking this PR as closed for now.

PR #83, opened 2020-07-13
Fix bad APIVersion for ClusterRoleBinding

lisa:
It should be rbac.authorization.k8s.io/v1, not v1.


lisa:
/hold

not sure why so much changed in the syncset

lisa:
/hold cancel

looks like there was a strange reorder. In a separate PR I'll sort the hooks by name.

PR #84, opened 2020-07-15
Initial effort at OSD-4046, protecting 3 additional namespaces

robotmaxtron:
Protecting the com, in, and io namespaces as part of OSD-4046

### DIFF
--- a/pkg/webhooks/namespace/namespace.go
+++ b/pkg/webhooks/namespace/namespace.go
@@ -19,7 +19,7 @@ import (
 
 const (
 	WebhookName                  string = "namespace-validation"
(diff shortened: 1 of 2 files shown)


lisa:
/hold

@robotmaxtron Could you squash this? (the hold is for the squash)

csheremeta:
Hey Max,

This looks good, but I think we could use more test cases - three additional cases stand out in my mind:

1. Ensuring that layered-sre-cluster-admins cannot create these namespaces either
2. Ensuring that clusterAdminUsers and sreAdminGroups can create these namespaces
3. Ensuring that the logic works as expected if the regex should not match, e.g. can dedicated-admins create namespaces like "common-apps" (matches ^com) or "test-admin" (matches in$)?

PR #85, opened 2020-07-16
Ensuring osd-impersonators group has regression tests

robotmaxtron:
Adding explicit testing for osd-impersonators group, see openshift/managed-cluster-config#390


lisa:
Let's keep this held til we have a good understanding of who is allowed to add users to this group. As mentioned in the related PR, there's some concern about Hive squashing membership that we'll want to be aware of anyways.

PR #87, opened 2020-07-20
OSD-4391: Allow certain @redhat.com users User accounts

lisa:
This PR changes many things:

First, in explanation of why so many webhooks are being changed: The test utils now supports for explicit oldObject `RawExtension` if the tester would so choose to provide one. They are optional, and the old behaviour is retained where they are not specified. (This also crosses off a TODO item.) This has been done to test a potential edge case where a user attempts to rename a User from unprivileged to privileged (`foo@example.com` to something @redhat.com).

More importantly, however, this adds a package (`userloader`) which supports the main impetus for this PR (see below): Only Red Hat associates who belong to certain groups may have User accounts on clusters. The group memberships are provided by a process outside the scope of this PR.

At the present time those groups are specified in the code as the `redhatGroups`, currently: `osd-devaccess`, `osd-sre-admins`, `layered-cs-sre-admins`. This PR does not care what kind of permissions those users may or may not have, only if they may have an account (which the permissions applied to those groups will apply).

In order to support this change, the ServiceAccount used by the oauth operator (`system:serviceaccount:openshift-authentication:oauth-openshift`) has been removed from the safelisted `kubeAdminUsernames` list because the Pod running with that SA is also (especially) required to follow these restrictions.

The `userloader` package is designed for use within a Kubernetes/OpenShift cluster. It uses the `rest.InClusterConfig()` to build up its connection. Because the operations supporting this PR are now taking place within the Webhook pods, the ServiceAccount has extended permissions (see `build/syncset.go`): `list` and `get` `users.openshift.io/groups`.

(Note: The related change to `build/selectorsyncset.yaml` has another subtle change for which a formal PR will come after this one: Webhook order is alphabetically sorted.)

Testing the `userloader` brings along the Go Mock framework, refer to the `userloader_test.go` file for usage.

Using gomock has an interface requirement and that interface serves us in testing the User webhook as well. The interface has a single defined method which the `userloader` itself implements.

```go
// This is coupled to Kubernetes
GetUsersFromGroups(...string) (map[string][]string, error)
```

The user hook's tests also (minimally) implements this interface:

```go
// empty for brevity
var testRedHatUsers = map[string][]string{}

// testUserLoader implements Loader
type testUserLoader struct{}

// GetUsersFromGroups implements userloader.Loader and is very minimal so as to
// isolate this user package from whatever the userloader package is doing
func (l *testUserLoader) GetUsersFromGroups(groups ...string) (map[string][]string, error) {
	return testRedHatUsers, nil
}

func testUserLoaderBuilder() (userloader.Loader, error) { return &testUserLoader{}, nil }

// makeTestHook sets up our fake data with our fake testUserLoader's
// implementation of userloader.Loader
func makeTestHook(t *testing.T) *UserWebhook {
	userLoaderBuilder = testUserLoaderBuilder
	return NewWebhook()
}
```

Group membership is queried for each well-formed request into the user webhook (cf `HandleRequest` method).

One question relating to this PR: Is it alright for group membership to be queried in this fasion? Should there be some explicit caching at this time (or later if it turns out to be an issue)?


lisa:
WIP is to address the final question and allow for in-cluster testing, but I want feedback before going too far down the rabbit path.

jaybeeunix:
how would a non-sre red hatter access their cluster? what's our "real" way of telling them to setup IdP, and do any of those give them @redhat.com usernames? (aka, does it need to consider the IdP in its check?)

lisa:
> how would a non-sre red hatter access their cluster? what's our "real" way of telling them to setup IdP, and do any of those give them @redhat.com usernames? (aka, does it need to consider the IdP in its check?)

@jaybeeunix Presumably they would fall into the `osd-devaccess` category (?) and in that case the associate would need to be added to that group before trying to log in.

jaybeeunix:
> > how would a non-sre red hatter access their cluster? what's our "real" way of telling them to setup IdP, and do any of those give them @redhat.com usernames? (aka, does it need to consider the IdP in its check?)
> 
> @jaybeeunix Presumably they would fall into the `osd-devaccess` category (?) and in that case the associate would need to be added to that group before trying to log in.

`osd-devaccess` is for cee. the case i'm talking about is if i'm a non-support/non-sre red hatter and i have an osd cluster, this would block my access if my IdP happened to use my redhat.com email as my username. i think it should only block @redhat.com if they're coming from OpenShift_SRE provider(s) and aren't in the approved groups, unless the typical way they'd access doesn't give them that kind of username.

lisa:
> > > how would a non-sre red hatter access their cluster? what's our "real" way of telling them to setup IdP, and do any of those give them @redhat.com usernames? (aka, does it need to consider the IdP in its check?)
> > 
> > 
> > @jaybeeunix Presumably they would fall into the `osd-devaccess` category (?) and in that case the associate would need to be added to that group before trying to log in.
> 
> `osd-devaccess` is for cee. the case i'm talking about is if i'm a non-support/non-sre red hatter and i have an osd cluster, this would block my access if my IdP happened to use my redhat.com email as my username. i think it should only block @redhat.com if they're coming from OpenShift_SRE provider(s) and aren't in the approved groups, unless the typical way they'd access doesn't give them that kind of username.

I think we would call that "working as intended" but I'll defer to @jewzaam for this one.

jewzaam:
Membership in a Group doesn't consider the Identity.  If a user were created with my email address in a non-Red Hat IdP they would get SRE access because the user's `name` is listed in the `osd-sre-admins` group.

Therefore creation of _any_ User with **@redhat.com** in the name must be gated.

jaybeeunix:
> Membership in a Group doesn't consider the Identity. If a user were created with my email address in a non-Red Hat IdP they would get SRE access because the user's `name` is listed in the `osd-sre-admins` group.
> 
> Therefore creation of _any_ User with **@redhat.com** in the name must be gated.

This change does not address that case, as noted [here in slack](https://coreos.slack.com/archives/CFJD1NZFT/p1595007624187800).

How does a non-SRE/non-CEE Red Hatter access their cluster when this change is in place? If we turn this code on as-is, everyone who isn't a SRE/CEE will have their access broken, even if they weren't using the "bad" IdP, if they happened to use their email address.

jewzaam:
Gotya, the creation of the user is gated, but since it's not tied to an identity provider it doesn't matter _where_ you try to create the user.  It will be allowed.  Is that an acceptable risk?  Only thing I can think to block that is also gate the creation of UserIdentityMapping resources.

The ID of Identity is well known, doesn't change between clusters (someone please verify before doing any work).  Logic would be:
* if identity is SRE's IdP and username not in allowed list:  DENY
* if identity is not(SRE's IdP) and username not in allowed list: ALLOW
* else: DENY

lisa:
> Gotya, the creation of the user is gated, but since it's not tied to an identity provider it doesn't matter _where_ you try to create the user. It will be allowed. Is that an acceptable risk? Only thing I can think to block that is also gate the creation of UserIdentityMapping resources.
> 
> The ID of Identity is well known, doesn't change between clusters (someone please verify before doing any work). Logic would be:
> 
> * if identity is SRE's IdP and username not in allowed list:  DENY
> * if identity is not(SRE's IdP) and username not in allowed list: ALLOW
> * else: DENY

@jewzaam To be clear it sounds like the scope of the ticket is changing thussly:

Premises:

1. safelist of SRE users is `[foo@example.com, bar@example.com]`
2. `example.com` users must follow IDP restrictions (follows)
3. the idp we expect SREs to use is `FooIDP` and the IDP for non-SRE Associates is `MyHttpIDP`, customer IDP `CustomerIDP`

Expected Behaviour:

1. A `User` being created for `foo@example.com` for idp `MyHttpIDP` is denied (expect this user to come from `FooIDP`
1. A `User` being created for non-SRE associate `alice@example.com` for idp `MyHttpIDP` is allowed (`User` is not in the safelist and so we expect them to *NOT* be using `FooIDP`)
1. A `User` is created `janet@example.net` using  `CustomerIDP` and they are not processed by this special case, as they're not an `example.com` protected domain
1. A `User` is created `janet@example.net` using `FooIDP` is denied (`FooIDP` is for protected users)
1. A `User` is created `janet@example.net` using `MyHttpIDP` is allowed (`MyHttpIDP` is for non-protected users)

I am not certain about the last case, but it would follow from the same reason for case 2.

jewzaam:
> Expected Behaviour:
> 
> 1. A User being created for foo@example.com for idp MyHttpIDP is denied (expect this user to come from FooIDP
> 2. A User being created for non-SRE associate alice@example.com for idp MyHttpIDP is allowed (User is not in the safelist and so we expect them to NOT be using FooIDP)
> 3. A User is created janet@example.net using CustomerIDP and they are not processed by this special case, as they're not an example.com protected domain
> 4. A User is created janet@example.net using FooIDP is denied (FooIDP is for protected users)
> 5. A User is created janet@example.net using MyHttpIDP is allowed (MyHttpIDP is for non-protected users)

Notes for each of the Expected Behaviors:
1. For clarity, it's denied because the user "foo@example.com" is in the safelist of SRE users.
2. +1
3. +1
4. This isn't a case we should care about because we control the SRE IdP, FooIDP.  All usernames for this IDP will be @example.com and nothing else.  Keep in mind we are going to throw this all away when we have backplane, the likelyhood of changes in associate email addresses eventually causing us to care about this case is a non-issue.
5. For clarity, we allow because it's not an "@example.com" user.

Rows 4 and 5 are really the same with the assumption that FooIDP users always have "@example.com" usernames.

jaybeeunix:
+1 to what @jewzaam said. Let me restate it to make sure we're all on the same page:

1. The safe list (priv'd) users must come from the SRE/FooIDP.
2. The only safe list users can come from FooIDP, all other users are denied if they attempt to use it.
3. Other than blocking safe list users from other IDPs, we don't care about them.

jaybeeunix:
The latest changes address my concerns and you have tests for them, so my vote is to continue progressing your work.

lisa:
/hold

for squash

lisa:
/hold cancel

squashed!

PR #89, opened 2020-07-22
Revert "OSD-4391: Allow certain @redhat.com users User accounts"

lisa:
Reverts openshift/managed-cluster-validating-webhooks#87

Running into issues in staging.

PR #90, opened 2020-07-22
WIP: OSD-4391: Fix oauth ServiceAccount check

lisa:
Previously the check used the username, and when the move to a
group-based check was made, that username was erronously used as the
groupname.

Will fix #87 and effectively revert #89.


lisa:
This branch is all kinds of messed up now. I'll fix it up in the morning.

pkg/webhooks/user/user.go
```diff
@@ -149,19 +230,59 @@ func (s *UserWebhook) HandleRequest(w http.ResponseWriter, r *http.Request) {
 		responsehelper.SendResponse(w, resp)
 		return
 	}
+	// load users in groups to know who is allowed an account
+	// TODO (lisa): Cache this and/or periodically refresh the data
+	if err = s.loadUsers(); err != nil {
```
jewzaam, line 235: This is possibly expensive.  Recommend move to where we really need it inside `isProtectedRedHatAssociate` to minimize calls out of the webhook to load data.

lisa:
Refer to #91

PR #93, opened 2020-07-31
OSD-4391: IDPs are prefixes

lisa:

### DIFF
--- a/pkg/webhooks/user/user.go
+++ b/pkg/webhooks/user/user.go
@@ -150,7 +150,7 @@ func (s *UserWebhook) isRedHatAssociate(userReq *userRequest) bool {
 
 func (s *UserWebhook) isUsingRedHatIDP(userReq *userRequest) bool {
 	for _, idp := range userReq.Identities {
-		if strings.HasPrefix(idp+ ":", redHatIDP) {
+		if strings.HasPrefix(idp+":", redHatIDP) {
 			return true
 		}
 	}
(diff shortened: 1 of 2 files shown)


pkg/webhooks/user/user.go
```diff
@@ -149,7 +149,12 @@ func (s *UserWebhook) isRedHatAssociate(userReq *userRequest) bool {
 }
 
 func (s *UserWebhook) isUsingRedHatIDP(userReq *userRequest) bool {
-	return utils.SliceContains(redHatIDP, userReq.Identities)
+	for _, idp := range userReq.Identities {
+		if strings.HasPrefix(idp, redHatIDP) {
```
lisa, line 153: Should this be `redHatIDP + ":"` since the format is `IDP:identifier`?

jewzaam, line 153: Given it's a prefix I recommend renaming args and the const to reflect that and to add the `:` into the prefix, else it will match something like `OpenShift_SRE_WasNotHere`.
https://github.com/openshift/managed-cluster-validating-webhooks/blob/909457725b7d2c177a049601c40f70fb93c6d2b3/pkg/webhooks/identity/identity.go#L21

jewzaam: Discussion on slack, sounds like a simple local update is best:

```suggestion
		if strings.HasPrefix(idp+ ":", redHatIDP) {
```

pkg/webhooks/user/user.go
```diff
@@ -149,7 +149,12 @@ func (s *UserWebhook) isRedHatAssociate(userReq *userRequest) bool {
 }
 
 func (s *UserWebhook) isUsingRedHatIDP(userReq *userRequest) bool {
-	return utils.SliceContains(redHatIDP, userReq.Identities)
+	for _, idp := range userReq.Identities {
+		if strings.HasPrefix(idp+ ":", redHatIDP) {
```
jewzaam, line 153:
```suggestion
		if strings.HasPrefix(idp+":", redHatIDP) {
```

PR #99, opened 2020-08-10
OSD-4749: Move #96 changes to code

lisa:
In #96 these were made in the YAML file, they are now moved to the code
file so that syncset.go will retain them in the future.


lisa:
This may also cause #98 to pass checks too.

PR #100, opened 2020-08-20
OSD-4884: admin group members can delete users with IDP mismatch

lisa:
Prior to the changes made in 92d168b446eb84b0284175164e524f483d092a46,
people could get a User account on cluster with an idp-group mismatch.
As part of the cleanup, we need to enable members of admin groups to
delete. This change fixes a bug with the cleanup


cblecker:
/lgtm
/hold

Holding if @jewzaam has any feedback. Can be removed at EOD if he doesn't have anything

PR #104, opened 2020-09-09
Webhook for blocking pods in namespaces openshift-logging, openshift-operators and non-core mnamespaces from deploying on master and infra.

aliceh:
This commit adds in pod.go a webhook that will check the namespace in which the pods deploy, and if that namespace is openshift-logging, openshift-operators or a non-core namespace the deployment will not be allowed on master and infra nodes. This commit prevents the logging stack from targeting the infra nodes [OSD-4262](https://issues.redhat.com/browse/OSD-5269) and prevents customers from launching pods with workloads on master and infra JIRA ticket [OSD-4333](https://issues.redhat.com/browse/OSD-4333).
The unit tests for the pod webhook are in pod_test.go.

### DIFF
--- a/pkg/webhooks/pod/pod.go
+++ b/pkg/webhooks/pod/pod.go
@@ -1,0 +1,187 @@
+package pod
+
+import (
+	"fmt"
+	"net/http"
+	"regexp"
+	"sync"
+
+	responsehelper "github.com/openshift/managed-cluster-validating-webhooks/pkg/helpers"
+	"github.com/openshift/managed-cluster-validating-webhooks/pkg/webhooks/utils"
+	"k8s.io/api/admission/v1beta1"
+	admissionregv1 "k8s.io/api/admissionregistration/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+	admissionctl "sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+const (
+	WebhookName           string = "pod-validation"
+	privilegedNamespace   string = `(^kube$|^kube-.*|^openshift$|^openshift-.*|^default$|^redhat-.*)`
+	unprivilegedNamespace string = `(openshift-logging|openshift-operators)`
+)
+
+var (
+	privilegedNamespaceRe   = regexp.MustCompile(privilegedNamespace)
+	unprivilegedNamespaceRe = regexp.MustCompile(unprivilegedNamespace)
+	log                     = logf.Log.WithName(WebhookName)
+
+	scope = admissionregv1.NamespacedScope
+	rules = []admissionregv1.RuleWithOperations{
+		{
+			Operations: []admissionregv1.OperationType{admissionregv1.OperationAll},
+			Rule: admissionregv1.Rule{
+				APIGroups:   []string{"v1"},
+				APIVersions: []string{"*"},
+				Resources:   []string{"pods"},
+				Scope:       &scope,
+			},
+		},
+	}
+)
+
+type PodWebhook struct {
+	mu sync.Mutex
+	s  runtime.Scheme
+}
+
+// TimeoutSeconds implements Webhook interface
+func (s *PodWebhook) TimeoutSeconds() int32 { return 1 }
+
+// MatchPolicy implements Webhook interface
+func (s *PodWebhook) MatchPolicy() admissionregv1.MatchPolicyType {
+	return admissionregv1.Equivalent
+}
+
+// Name implements Webhook interface
+func (s *PodWebhook) Name() string { return WebhookName }
+
+// FailurePolicy implements Webhook interface and defines how unrecognized errors and timeout errors from the admission webhook are handled. Allowed values are Ignore or Fail.
+// Ignore means that an error calling the webhook is ignored and the API request is allowed to continue.
+// It's important to leave the FailurePolicy set to Ignore because otherwise the pod will fail to be created as the API request will be rejected.
+func (s *PodWebhook) FailurePolicy() admissionregv1.FailurePolicyType {
+	return admissionregv1.Ignore
+}
+
+// Rules implements Webhook interface
+func (s *PodWebhook) Rules() []admissionregv1.RuleWithOperations { return rules }
+
+// GetURI implements Webhook interface
+func (s *PodWebhook) GetURI() string { return "/" + WebhookName }
+
+// SideEffects implements Webhook interface
+func (s *PodWebhook) SideEffects() admissionregv1.SideEffectClass {
+	return admissionregv1.SideEffectClassNone
+}
+
+// Validate implements Webhook interface
+func (s *PodWebhook) Validate(req admissionctl.Request) bool {
+	valid := true
+	valid = valid && (req.UserInfo.Username != "")
+	valid = valid && (req.Kind.Kind == "Pod")
+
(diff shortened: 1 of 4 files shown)


lisa:
/hold


for review

pkg/webhooks/pod/pod.go
```diff
@@ -0,0 +1,174 @@
...
+	return admissionregv1.Equivalent
+}
+
+// Name implements Webhook interface
+func (s *PodWebhook) Name() string { return WebhookName }
+
+// FailurePolicy implements Webhook interface
+func (s *PodWebhook) FailurePolicy() admissionregv1.FailurePolicyType {
+	return admissionregv1.Ignore
+}
```
cblecker, line 63: This should have a comment about why we are ignoring it, and how important it is that this doesn't change.

lisa (changes requested):
Hey @aliceh great start! You're missing a way to register the webhook, look at the other `add_*.go` files for inspiration. Once that file is in place you need to `make syncset` and commit the new `build/selectorsyncset.yaml` file.

I'd also like to see a test case for infra node and noschedule taint, please.

/cc @cblecker as well.

pkg/webhooks/pod/pod_test.go
```diff
@@ -0,0 +1,482 @@
...
+				{
+					Key:      "node-role.kubernetes.io/infra",
+					Operator: corev1.TolerationOpEqual,
+					Value:    "toleration key value2",
+					Effect:   corev1.TaintEffectNoSchedule,
+				},
+			},
+			operation:       v1beta1.Create,
+			shouldBeAllowed: false,
+		},
```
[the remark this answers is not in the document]
aliceh: my understanding was that the non-core namespaces should be blocked from deploying pods on master/infra when there's toleration for NoSchedule taint.

pkg/webhooks/pod/pod.go
```diff
@@ -0,0 +1,174 @@
...
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+	admissionctl "sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+const (
+	WebhookName         string = "pod-validation"
+	privilegedNamespace string = `(^kube$|^kube-.*|^openshift$|^openshift-.*|^default$|^redhat-.*)`
+	exceptionNamespace  string = `(openshift-logging|openshift-operators)`
```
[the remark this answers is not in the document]
aliceh: The privileged namespaces are the ones that are able to deploy, we are excluding from that set the exception namespaces openshift-logging and openshift-operators which we block from deploying, same as non-privileged customer namespaces.

pkg/webhooks/pod/pod.go
```diff
@@ -0,0 +1,174 @@
...
+
+func (s *PodWebhook) authorized(request admissionctl.Request) admissionctl.Response {
+	var ret admissionctl.Response
+	pod, err := s.renderPod(request)
+	if err != nil {
+		log.Error(err, "Couldn't render a Pod from the incoming request")
+		return admissionctl.Errored(http.StatusBadRequest, err)
+	}
+	//If the incoming Pod is aimed at a privileged namespace except for openshift-logging and openshift-operators, allow it to do whatever it wants.
+	//However, if the pod is targeting a customer's namespace (aka non-privileged), then it may not tolerate certain master/infra node taints.
```
lisa, line 112:
```suggestion
	// If the incoming Pod is aimed at a privileged namespace except for exceptionNamespace, allow it to do whatever it wants.
	// However, if the pod is targeting a customer's namespace (aka non-privileged), then it may not tolerate certain master/infra node taints.
```

pkg/webhooks/pod/pod.go
```diff
@@ -0,0 +1,174 @@
...
+	if !privilegedNamespaceRe.Match([]byte(pod.ObjectMeta.GetNamespace())) || exceptionNamespaceRe.Match([]byte(pod.ObjectMeta.GetNamespace())) {
```
lisa, line 113: This is a fairly complex conditional and it'd be clearer to express what is being matched if it were refactored into a well-named function. Perhaps `isRequestPrivileged` or `isRequestUnprivileged`.

pkg/webhooks/pod/pod.go
```diff
@@ -0,0 +1,174 @@
...
+			if toleration.Key == "node-role.kubernetes.io/master" && toleration.Effect == corev1.TaintEffectPreferNoSchedule {
+				ret = admissionctl.Denied("Not allowed to schedule a pod with PreferNoSchedule taint on master node")
+				ret.UID = request.AdmissionRequest.UID
+				return ret
+			}
+		}
+	}
+
+	// Hereafter, all requests are controlled by RBAC
+	ret = admissionctl.Allowed("Allowed to create a Pod in a privileged Namespace because of RBAC")
```
lisa, line 139:
```suggestion
	ret = admissionctl.Allowed("Allowed to create Pod because of RBAC")
```

Here it will apply for all namespaces, not just privileged ones.

pkg/webhooks/pod/pod_test.go
```diff
@@ -0,0 +1,542 @@
...
+
+	"github.com/openshift/managed-cluster-validating-webhooks/pkg/testutils"
+	"k8s.io/api/admission/v1beta1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+func createRawPodJSON(name string, tolerations []corev1.Toleration, uid string, namespace string) (string, error) {
+
```
lisa, line 16:
```suggestion
```

Extra newline

pkg/webhooks/pod/pod.go
```diff
@@ -0,0 +1,174 @@
...
+	ret.UID = request.AdmissionRequest.UID
+	return ret
+}
+
+// HandleRequest Decide if the incoming request is allowed
```
lisa, line 144: There will need to be some refactor once #105 is merged.

lisa (changes requested):
Some feedback.

I'll continue with unit test review, but my day is nearly over and I wanted to get this to you in time.

pkg/webhooks/pod/pod.go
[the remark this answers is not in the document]
aliceh: My reasoning behind the choice of name was that we have the core namespaces that are privileged as a group in other contexts too for example in the case of RBAC subject permissions, so the complement of that group would be the unprivileged namespaces. In this context, it seemed reasonable to call the set of two namespaces that we want to exclude from the privileged set exceptions. Of course, I am happy to rename that set unprivileged if you feel that it's better.

lisa:
@cblecker Looks okay to me. What are your thoughts?

@aliceh Can you squash the commits into one?

lisa:
/lgtm
/hold

Please let @csheremeta or @cblecker remove my hold since I am out for four days and #109 needs to take priority for promoting.

lisa:
@aliceh @cblecker Is this ready to merge still?

jewzaam:
Only going by the title, but is this not limiting Projects from deploying to master / infra?

jewzaam:
> Only going by the title, but is this not limiting Projects from deploying to master / infra?

I didn't notice the "non-core"  mention in the title , my bad.

lisa (approved):

PR #109, opened 2020-10-08
Initial attempt at subscription webhook to block logging 4.5 installs…

csheremeta:
…/upgrades
https://issues.redhat.com/browse/OSD-5469
/hold for testing


pkg/webhooks/subscription/subscription.go
```diff
@@ -0,0 +1,182 @@
...
+
+// isLogging45Request is a helper to consolidate logic. Returns true
+// if the request is for the cluster-logging or elasticsearch-operator
+// subscription on 4.5 channel
+func (s *SubscriptionWebhook) isLogging45Request(subscriptionReq *subscriptionRequest) bool {
+	if subscriptionReq.Spec.Name == LoggingSubName || subscriptionReq.Spec.Name == ESSubName {
+		if subscriptionReq.Spec.Channel == Channel {
+			return true
+		}
+	}
```
lisa, line 105:
```suggestion
	if subscriptionReq.Spec.Channel == Channel && (subscriptionReq.Spec.Name == LoggingSubName || subscriptionReq.Spec.Name == ESSubName) {
		return true
	}
```

This could work a little better? Shortcutting on the channel check at first.

pkg/webhooks/subscription/subscription.go
```diff
@@ -0,0 +1,182 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+	admissionctl "sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+const (
+	WebhookName    string = "subscription-validation"
+	LoggingSubName string = "cluster-logging"
+	ESSubName      string = "elasticsearch-operator"
+	Channel        string = "4.5"
```
lisa, line 21: What does `Channel` signify? Is it a channel that's blocked somehow?

pkg/webhooks/subscription/subscription.go
```diff
@@ -0,0 +1,182 @@
...
+)
+
+var (
+	privilegedUsers = []string{"kube:admin", "system:admin", "system:serviceaccount:openshift-authentication:oauth-openshift"}
```
lisa, line 25: > `system:serviceaccount:openshift-authentication:oauth-openshift`

Does the OAUTH operator interact with Subscription objects?

lisa (changes requested):
Couple questions, minor feedback on the code here.

We also need some tests.

fahlmant (commented):
Overall seems straightforward to me it terns of logic

pkg/webhooks/subscription/subscription.go
csheremeta: Yes, 4.5 is the channel we want to block from being installed or upgraded to

lisa: May I suggest changing it to `BlockedChannel` or something like that?

pkg/webhooks/subscription/subscription.go
csheremeta: I'm not sure. I ripped the privilegedUsers var out of some other webhook to save time. I'm ok to rip that bit out. I think there may be some service account we want to account for, but I need to do some manual testing to figure that out

pkg/webhooks/subscription/subscription.go
csheremeta: Sure :) my concern was readability, but I can add comments

pkg/webhooks/subscription/subscription.go
csheremeta: +1

pkg/webhooks/subscription/subscription.go
lisa: That's actually a good question. I'm not sure what's responsible for updating this in-cluster. My first guess would be OLM.

pkg/webhooks/subscription/subscription.go
```diff
@@ -0,0 +1,183 @@
...
+const (
+	WebhookName    string = "subscription-validation"
+	LoggingSubName string = "cluster-logging"
+	ESSubName      string = "elasticsearch-operator"
+	BlockedChannel string = "4.5"
+)
+
+var (
+	privilegedUsers = []string{"kube:admin", "system:admin"}
+	adminGroups     = []string{"osd-sre-admins", "osd-sre-cluster-admins"}
```
lisa, line 26: What if we took a different approach with this hook? We know that  customers will have specific group membership and so instead of saying that members of these lists are _allowed_, we could say that members of those customer groups are _disallowed_ and everyone else is allowed.

The default case will be (RBAC-based) allow unless the requestor is a member of a customer group trying to make changes.

pkg/webhooks/subscription/subscription.go
```diff
@@ -0,0 +1,183 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+	admissionctl "sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+const (
+	WebhookName    string = "subscription-validation"
+	LoggingSubName string = "cluster-logging"
+	ESSubName      string = "elasticsearch-operator"
+	BlockedChannel string = "4.5"
```
rogbas, line 21: Should we make this an array? Only to make it easier to do last minute changes in case we need to block 4.6 channel as well

lisa: I'd +1 this if we really expect to need to do this. My assumption is that this hook will be relatively short-lived.

pkg/webhooks/subscription/subscription.go
```diff
@@ -0,0 +1,183 @@
...
+		if utils.SliceContains(group, adminGroups) {
+			ret = admissionctl.Allowed("Members of admin group may install or upgrade logging 4.5 operator")
+			ret.UID = request.AdmissionRequest.UID
+			return ret
+		}
+	}
+
+	// if we're here, non-privileged user is attempting to CREATE or UPDATE logging
+	// operator at 4.5 - deny this
+	ret = admissionctl.Denied("Non-privileged users cannot install or upgrade logging 4.5 operator")
```
rogbas, line 151: Do we want to be more straight forward here, stating there are known issues with the logging-operator v4.5?

rogbas:
@csheremeta The logic looks good, I just left 2 minor comments.

pkg/webhooks/subscription/subscription.go
rogbas: IMO it's simpler to manage access the way it's coded now, allowing only privileged users to do it. 

Doing the opposite and Identifying customer's groups seems harder to implement.

pkg/webhooks/subscription/subscription.go
rogbas: +1, it makes sense.

It seems 4.6 goes GA in 2 weeks. I don't know the timeline for the logging-team solving the open issues.

pkg/webhooks/subscription/subscription.go
csheremeta: FTR, turns out the SA we needed to add was system:serviceaccount:kube-system:generic-garbage-collector. This allows the logging stack to be deleted

pkg/webhooks/subscription/subscription.go
csheremeta: :+1:

pkg/webhooks/subscription/subscription.go
csheremeta: I'm ok either way here. it's easy to implement. however, without being able to upgrade to 4.5, customers won't be able to get from 4.4 to 4.6 without ripping out the logging stack and reinstalling. and considering we're not sure if the issues are seen only when upgrading from 4.4 to 4.5 (or if they're also seen with fresh installs of 4.5), it's possible that ripping out 4.4 and reinstalling at 4.6 won't be problematic, if that's what the customer wants to do.

either way, I also hope this is short-lived and won't live long enough to see a need for this

pkg/webhooks/subscription/subscription.go
csheremeta: yes. I'll add in more info here

pkg/webhooks/subscription/subscription.go
```diff
@@ -22,8 +22,8 @@ const (
 )
 
 var (
-	privilegedUsers = []string{"kube:admin", "system:admin"}
-	adminGroups     = []string{"osd-sre-admins", "osd-sre-cluster-admins"}
+	privilegedUsers = []string{"kube:admin", "system:admin", "system:serviceaccount:kube-system:generic-garbage-collector"}
```
lisa, line 25:
```suggestion
	privilegedUsers = []string{"kube:admin", "system:admin"}
```
This is redundant with the next line because the GC SA is a member of the `system:serviceaccount:kube-system` group.

pkg/webhooks/subscription/subscription.go
csheremeta: something like ""Non-privileged users cannot install or upgrade to the v4.5 logging operator at this time, as there are known issues with logging v4.5 which we are working to resolve." ?

pkg/webhooks/subscription/subscription.go
csheremeta: after some manual testing, I think my vote would be to leave the generic-garbage-collector in the user list and remove kube-system from the groups

lisa: Either way works for me.

csheremeta:
manual testing and unit tests all pass
/unhold

pkg/webhooks/subscription/subscription.go
rogbas: +1 to keeping a single blockedChannel for now

pkg/webhooks/subscription/subscription.go
```diff
@@ -0,0 +1,183 @@
...
+// FailurePolicy implements Webhook interface
+func (s *SubscriptionWebhook) FailurePolicy() admissionregv1.FailurePolicyType {
+	return admissionregv1.Ignore
+}
+
+// Rules implements Webhook interface
+func (s *SubscriptionWebhook) Rules() []admissionregv1.RuleWithOperations { return rules }
+
+// GetURI implements Webhook interface
+func (s *SubscriptionWebhook) GetURI() string { return "/subscription-validation" }
```
lisa, line 81:
```suggestion
func (s *SubscriptionWebhook) GetURI() string { return "/" + WebhookName }
```

Sorry for the late find. Can we do this instead please?

lisa (changes requested):

lisa (approved):
/lgtm

PR #113, opened 2020-10-20
OSD-5377 Fine tune error messages for namespace webhook

csheremeta:
See https://docs.google.com/spreadsheets/d/1sT3loQphcc75k3W3EKCcyeSTQ1ppK7Md7ASB0CUSOLQ/edit#gid=0

### DIFF
--- a/pkg/webhooks/namespace/namespace.go
+++ b/pkg/webhooks/namespace/namespace.go
@@ -19,7 +19,8 @@ import (
 
 const (
 	WebhookName                  string = "namespace-validation"
-	privilegedNamespace          string = `(^kube.*|^openshift.*|^default$|^redhat.*|^com$|^io$|^in$)`
+	privilegedNamespace          string = `(^kube.*|^openshift.*|^default$|^redhat.*)`
+	badNamespace                 string = `(^com$|^io$|^in$)`
 	privilegedServiceAccounts    string = `^system:serviceaccounts:(kube.*|openshift.*|default|redhat.*)`
 	layeredProductNamespace      string = `^redhat.*`
 	layeredProductAdminGroupName string = "layered-sre-cluster-admins"
@@ -30,6 +31,7 @@ var (
 	sreAdminGroups    = []string{"osd-sre-admins", "osd-sre-cluster-admins"}
 
 	privilegedNamespaceRe       = regexp.MustCompile(privilegedNamespace)
+	badNamespaceRe              = regexp.MustCompile(badNamespace)
 	privilegedServiceAccountsRe = regexp.MustCompile(privilegedServiceAccounts)
 	layeredProductNamespaceRe   = regexp.MustCompile(layeredProductNamespace)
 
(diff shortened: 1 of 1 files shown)


pkg/webhooks/namespace/namespace.go
```diff
@@ -151,7 +153,28 @@ func (s *NamespaceWebhook) authorized(request admissionctl.Request) admissionctl
...
+				amISREAdmin = true
+				break
+			}
+		}
+		if amIClusterAdmin || amISREAdmin {
+			ret = admissionctl.Allowed("Cluster and SRE admins may access")
+			ret.UID = request.AdmissionRequest.UID
+			return ret
+		}
+		log.Info("Non-admin access attempt to privileged namespace", "request", request.AdmissionRequest)
```
lisa, line 176: We can probably change this info message to be specific to DNS.

csheremeta: good catch, fixed

pkg/webhooks/namespace/namespace.go
```diff
@@ -150,8 +152,29 @@ func (s *NamespaceWebhook) authorized(request admissionctl.Request) admissionctl
...
+				break
+			}
+		}
+		if amIClusterAdmin || amISREAdmin {
+			ret = admissionctl.Allowed("Cluster and SRE admins may access")
+			ret.UID = request.AdmissionRequest.UID
+			return ret
+		}
+		log.Info("Non-admin (not 'kube:admin' or 'system:admin', or a member of osd-sre-admins or osd-sre-cluster-admins) attempted to access a potentially harmful namespace (eg matching this regex", "regex", badNamespace, "request", request.AdmissionRequest)
+		ret = admissionctl.Denied(fmt.Sprintf("Prevented from creating a potentially harmful namespace. Customer namespaces should not begin with 'com', 'io', or 'in' as this would impact DNS resolution."))
```
lisa, line 177: With the log line I think we can omit the list of groups here because we'll know what code is in use and what safelisted groups are, so it's extra stuff.

> Customer namespaces should not begin with 'com', 'io', or 'in' as this would impact DNS resolution.

Check the regular expression :) It's `^com$` (etc) which means the name can't be exactly `com` (etc). May want to provide the regex here as well, like with the previous check.

csheremeta: +1. fixed again

fahlmant (approved):
/lgtm

pkg/webhooks/namespace/namespace.go
```diff
@@ -150,8 +152,29 @@ func (s *NamespaceWebhook) authorized(request admissionctl.Request) admissionctl
 			ret.UID = request.AdmissionRequest.UID
 			return ret
 		}
-		log.Info("Non-admin access attempt to privileged namespace", "request", request.AdmissionRequest)
-		ret = admissionctl.Denied(fmt.Sprintf("Prevented from creating or changing a potentially harmful Namespace. Customer namespaces should NOT match this regular expressions: %s", privilegedNamespace))
+		log.Info("Non-admin attempted to access a privileged namespace (eg matching this regex)", "regex", privilegedNamespace, "request", request.AdmissionRequest)
+		ret = admissionctl.Denied(fmt.Sprintf("Prevented from accessing Red Hat managed namespaces. Customer workloads should be placed in customer namespaces, and should not match this regex: %s", privilegedNamespace))
```
lisa, line 156:
```suggestion
		ret = admissionctl.Denied(fmt.Sprintf("Prevented from accessing Red Hat managed namespaces. Customer workloads should be placed in customer namespaces, and should not match this regular expression: %s", privilegedNamespace))
```

To be consistent with "regular expression" in the other hooks, as opposed to "regex".

pkg/webhooks/namespace/namespace.go
```diff
@@ -150,8 +152,29 @@ func (s *NamespaceWebhook) authorized(request admissionctl.Request) admissionctl
...
+				break
+			}
+		}
+		if amIClusterAdmin || amISREAdmin {
+			ret = admissionctl.Allowed("Cluster and SRE admins may access")
+			ret.UID = request.AdmissionRequest.UID
+			return ret
+		}
+		log.Info("Non-admin attempted to access a potentially harmful namespace (eg matching this regex)", "regex", badNamespace, "request", request.AdmissionRequest)
+		ret = admissionctl.Denied(fmt.Sprintf("Prevented from creating a potentially harmful namespace. Customer namespaces should not match this regex, as this would impact DNS resolution: %s", badNamespace))
```
lisa, line 177:
```suggestion
		ret = admissionctl.Denied(fmt.Sprintf("Prevented from creating a potentially harmful namespace. Customer namespaces should not match this regular expression, as this would impact DNS resolution: %s", badNamespace))
```

Same as above.

lisa (commented):
minor wording

pkg/webhooks/namespace/namespace.go
csheremeta: done :)

PR #114, opened 2020-10-20
Split group checks between admin and managed groups

fahlmant:
Reference: https://issues.redhat.com/browse/OSD-5378

This is my first pass at understanding this repo, so please let me know if I'm completely misunderstanding. 

What's the relation between this PR and this sheet? https://docs.google.com/spreadsheets/d/1sT3loQphcc75k3W3EKCcyeSTQ1ppK7Md7ASB0CUSOLQ/edit#gid=0

Should I be using the error messages from there instead? 

Another small question, what should be done with `osd-devaccess`?

### DIFF
--- a/pkg/webhooks/group/group.go
+++ b/pkg/webhooks/group/group.go
@@ -36,15 +36,17 @@ type groupRequest struct {
 }
 
 const (
-	WebhookName     string = "group-validation"
-	protectedGroups string = `(^osd.*|^dedicated-admins$|^cluster-admins$|^layered-cs-sre-admins$)`
+	WebhookName               string = "group-validation"
+	protectedAdminGroups      string = `(^dedicated-admins$|^cluster-admins$)`
+	protectedManagementGroups string = `(^osd-sre-admins$|^osd-sre-cluster-admins$|^osd-devaccess$|^layered-cs-sre-admins$)`
 )
 
 var (
-	log               = logf.Log.WithName(WebhookName)
-	protectedGroupsRe = regexp.MustCompile(protectedGroups)
-	clusterAdminUsers = []string{"kube:admin", "system:admin"}
-	adminGroups       = []string{"osd-sre-admins", "osd-sre-cluster-admins"}
+	log                         = logf.Log.WithName(WebhookName)
+	protectedAdminGroupsRe      = regexp.MustCompile(protectedAdminGroups)
+	protectedManagementGroupsRe = regexp.MustCompile(protectedManagementGroups)
+	clusterAdminUsers           = []string{"kube:admin", "system:admin"}
+	adminGroups                 = []string{"osd-sre-admins", "osd-sre-cluster-admins"}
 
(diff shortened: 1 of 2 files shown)


csheremeta:
> What's the relation between this PR and this sheet?

The rows that are highlight in green are the changes requested for 5377 and 5378, although rows 4 and 6 also represent changes from the current state (you'll want to focus on 5378, or rows 6 and 7). If you click on the "OSD-5226" tab in the spreadsheet, you'll see the current state of the error messages

> Should I be using the error messages from there instead?

Yes, although we can modify them, and we probably should for this case. I also modified the ones I used for 5377 - see the "Changed To" column on that sheet

> Another small question, what should be done with `osd-devaccess`?

This should be included with the `protectedManagementGroups`, as customers should never be able to modify this group (by contrast, the groups in `protectedAdminGroups` can be modified by customers, but should be done through OCM)

pkg/webhooks/group/group.go
```diff
@@ -36,15 +36,17 @@ type groupRequest struct {
 }
 
 const (
-	WebhookName     string = "group-validation"
-	protectedGroups string = `(^osd.*|^dedicated-admins$|^cluster-admins$|^layered-cs-sre-admins$)`
+	WebhookName               string = "group-validation"
+	protectedAdminGroups      string = `(^dedicated-admins$|^cluster-admins$)`
+	protectedManagementGroups string = `(^osd.*|^layered-cs-sre-admins$)`
```
csheremeta, line 41:
```suggestion
	protectedManagementGroups string = `(^osd-sre-admins$|^osd-sre-cluster-admins$|^osd-devaccess$|^layered-cs-sre-admins$)`
```

I think we should hard-code the 3 specific groups here, so that customers can create other groups that start with osd if they want to

fahlmant: :+1: to this. I saw in the ticket there should be conversation about it, so I figured PR is where we should

pkg/webhooks/group/group.go
```diff
@@ -114,11 +117,28 @@ func (s *GroupWebhook) authorized(request admissionctl.Request) admissionctl.Res
 				return ret
 			}
 		}
+
+		log.Info("Denying access", "request", request.AdmissionRequest)
+		ret = admissionctl.Denied(fmt.Sprintf("Prevented from accessing Red Hat managed Admin groups. Customer user groups may NOT match this regular expression: %s", protectedAdminGroups))
```
csheremeta, line 122:
```suggestion
		ret = admissionctl.Denied(fmt.Sprintf("Prevented from accessing Red Hat managed Admin groups. Customers should use https://cloud.redhat.com/openshift to manage groups that match this regular expression: %s", protectedAdminGroups))
```

This is a bit of a modification to the proposed message from the spreadsheet. @lisa will probably want to vet the specific messages

lisa: 👍  to Candace's suggested change here.

PR #116, opened 2020-10-23
Update error message for operations on default identity - OSD-5379

csheremeta:


pkg/webhooks/identity/identity.go
```diff
@@ -121,7 +121,7 @@ func (s *IdentityWebhook) authorized(request admissionctl.Request) admissionctl.
 			}
 		}
 		log.Info("Denying access", "request", request.AdmissionRequest)
-		ret = admissionctl.Denied("Prevented from accessing Red Hat managed identity provider. Please manage your identity provider through https://cloud.redhat.com/openshift")
+		ret = admissionctl.Denied("Prevented from modifying Red Hat's managed Identity. You may create/modify any Identity objects, except for 'OpenShift_SRE'.")
```
lisa, line 124: Should reference the constant in this package.

lisa (changes requested):
/hold

pkg/webhooks/identity/identity.go
lisa, line 124:
```suggestion
		ret = admissionctl.Denied(fmt.Sprintf("Prevented from modifying Red Hat's managed Identity. You may create/modify any Identity objects, except for %s.", DefaultIdentityProvider))
```

lisa: Suggestion to implement for my [previous comment](https://github.com/openshift/managed-cluster-validating-webhooks/pull/116/files#r511300954).

lisa (changes requested):

csheremeta:
@lisa suggested changes have been committed, let me know if you think this looks good to go

lisa:
/unhold

whoops, forgot one important step

PR #105, opened 2020-09-15
DRY out HTTP request handling

lisa:
This commit will remove the `HandleRequest` requirement and formalize
the heretofore informal `authorized` methods into the Webhook interface
by introducing a new dispatcher package that is responsible for parsing
the HTTP payload into a Request object that webhookscan deal more
natively with. Each hook, with this change, will shed any knowledge of
the transport mechanism and only work with the higher level logic
specifically around webhooks.

Also updated are documentation and testutils package to support these
new changes.

Misc bug fixes:

* HTTP Status is correctly written copied from the response object
(unless it was omitted from the object somehow) * Fixed some copy/paste
errors

Signed-off-by: Lisa Seelye <lisa@users.noreply.github.com>

fyi @cblecker @aliceh since if merged this will impact your work in #104 

/kind cleanup

PR #117, opened 2020-11-10
[WIP] New nodeLabels webhook

ArielLima:
OpenShift Dedicated customers should be allowed to edit node labels using dedicated-admin and cluster-admin.

- Allow dedicated-admin to apply node labels in RBAC
- Update webhook to not allow master/infra nodes from being edited.
- Update webhook to not allow worker nodes from losing the "worker" label.  This is needed for OCP/OCM/HIVE.
- Update webhook so no new nodes may have the master or infra label.

Card -> https://issues.redhat.com/browse/OSD-5225


lisa:
Recall that `regularuser` also hooks some node things, now is probably a good time to split that out and these concerns into their own specific node webhook.

/hold

(for future review once wip is done)

lisa:
@ArielLima Where does this stand?

Note that it will need a rebase with the merging of #105 (the README.md has been updated).

pkg/webhooks/node/node.go
```diff
@@ -0,0 +1,230 @@
...
+}
+
+func contains(s []string, e string) bool {
+	for _, a := range s {
+		if a == e {
+			return true
+		}
+	}
+	return false
+}
```
lisa, line 196: Look into `utils.SliceContains`

rogbas:
This PR is on hold, waiting on on an upstream issue in k8s to be addressed. In March 2021 we'll revisit it.

lisa:
This work is waiting on https://github.com/kubernetes/kubernetes/issues/100096

rogbas:
@rogbas update PR with findings from the respective k8s CVE

PR #118, opened 2020-11-11
Block logging 4.6 installs/upgrades through subscription webhook

csheremeta:


pkg/webhooks/subscription/subscription.go
```diff
@@ -94,11 +94,11 @@ func (s *SubscriptionWebhook) Validate(req admissionctl.Request) bool {
...
 }
 
-// isLogging45Request is a helper to consolidate logic. Returns true
+// isLogging45or46Request is a helper to consolidate logic. Returns true
 // if the request is for the cluster-logging or elasticsearch-operator
-// subscription on 4.5 channel
-func (s *SubscriptionWebhook) isLogging45Request(subscriptionReq *subscriptionRequest) bool {
-	if subscriptionReq.Spec.Channel == BlockedChannel && (subscriptionReq.Spec.Name == LoggingSubName || subscriptionReq.Spec.Name == ESSubName) {
+// subscription on 4.5 or 4.6 channels
+func (s *SubscriptionWebhook) isLogging45or46Request(subscriptionReq *subscriptionRequest) bool {
```
lisa, line 100: `isLogging45or46Request` to `isBlockedLoggingUpgradeRequest`?

(Although this makes me miss Ruby's `?` methods: `is_upgrade_blocked?` would be so nice.)

csheremeta: How do you feel about `isBlockedLoggingRequest`? We're blocking installs too (not just upgrades)

lisa: or `isLoggingRequestBlocked` to be read a bit more English-like? (hello COBOL)

csheremeta: I stuck with `isBlockedLoggingRequest` because in my mind, "Is it a blocked logging request?" is the better question over "Is the logging request blocked?", since the second question presumes that the request is a logging request, when we don't actually know that until we get into the function itself.

Semantics to the extreme. Let me know if you're good with this :)

fahlmant (approved):
/lgtm

PR #125, opened 2020-12-14
OSD-2019: Allow dedicated-admins to label namespaces

lisa:
Allow dedicated-admins to label customer namespaces while still
protecting quota-related labels (See
https://github.com/openshift/managed-cluster-config/tree/master/deploy/resource-quotas)

### DIFF
--- a/pkg/webhooks/namespace/namespace.go
+++ b/pkg/webhooks/namespace/namespace.go
@@ -23,7 +23,7 @@ const (
 	privilegedServiceAccounts    string = `^system:serviceaccounts:(kube.*|openshift.*|default|redhat.*|osde2e-[a-z0-9]{5})`
 	layeredProductNamespace      string = `^redhat.*`
 	layeredProductAdminGroupName string = "layered-sre-cluster-admins"
-	docString                    string = `Managed OpenShift Customers may not modify privileged namespaces identified by this regular expression %s because customer workloads should be placed in customer-created namespaces. Customers may not create namespaces identified by this regular expression %s because it could interfere with critical DNS resolution.`
+	docString                    string = `Managed OpenShift Customers may not modify privileged namespaces identified by this regular expression %s because customer workloads should be placed in customer-created namespaces. Customers may not create namespaces identified by this regular expression %s because it could interfere with critical DNS resolution. Additionally, customers may not set or change the values of these Namespace labels %s.`
 )
 
 var (
@@ -34,6 +34,13 @@ var (
 	badNamespaceRe              = regexp.MustCompile(badNamespace)
 	privilegedServiceAccountsRe = regexp.MustCompile(privilegedServiceAccounts)
 	layeredProductNamespaceRe   = regexp.MustCompile(layeredProductNamespace)
+	// protectedLabels are labels which managed customers should not be allowed
+	// change by dedicated-admins.
+	protectedLabels = []string{
+		// https://github.com/openshift/managed-cluster-config/tree/master/deploy/resource-quotas
+		"managed.openshift.io/storage-pv-quota-exempt",
+		"managed.openshift.io/service-lb-quota-exempt",
+	}
 
 	log = logf.Log.WithName(WebhookName)
 
@@ -58,7 +65,7 @@ type NamespaceWebhook struct {
 }
 
 func (s *NamespaceWebhook) Doc() string {
-	return fmt.Sprintf(docString, privilegedNamespace, badNamespace)
+	return fmt.Sprintf(docString, privilegedNamespace, badNamespace, protectedLabels)
 }
 
 // TimeoutSeconds implements Webhook interface
@@ -97,7 +104,20 @@ func (s *NamespaceWebhook) Validate(req admissionctl.Request) bool {
 	return valid
 }
 
-// renderNamespace pluck out the Namespace from the Object or OldObject
+// renderNamespace decodes a *corev1.Namespace from the incoming request and
+// gives preference to the OldObject (if it exists) over the Object. This method
+// is functionally similar to the renderOldAndNewNamespaces method except we
+// want to use this method when the assertions do not necessarily care which
+// verb is being performed. That is, if the assertion is for a CREATE we want to
+// use the request.Object, if it is for an UPDATE verb, we want to reference
+// what the object was prior to the change (request.OldObject). This view can be
+// seen similar to an "authenticated" check because the output of this method is
+// used to determine if the requestor can make any kind of change whatsoever
+// (ignoring the particulars of the change). Later, in unauthorizedLabelChanges,
+// we use renderOldAndNewNamespaces to take into account the particulars of the
+// kinds of changes we care about, with the current (request.Object) and former
+// (request.OldObject) objects returned. See the renderOldAndNewNamespaces
+// documentation for more.
 func (s *NamespaceWebhook) renderNamespace(req admissionctl.Request) (*corev1.Namespace, error) {
 	decoder, err := admissionctl.NewDecoder(&s.s)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/webhooks/namespace/namespace.go
```diff
@@ -97,7 +104,11 @@ func (s *NamespaceWebhook) Validate(req admissionctl.Request) bool {
 	return valid
 }
 
-// renderNamespace pluck out the Namespace from the Object or OldObject
+// renderNamespace decodes a *corev1.Namespace from the incoming request and
+// gives preference to the OldObject (if it exists) over the Object. This
+// preference favours the Create and Update operations but falls short of fully
```
csheremeta, line 109:
```suggestion
// preference favours the Create and Delete operations but falls short of fully
```
I think this is perhaps a little misleading, in that it implies that we use this function to complete the webhook's functionality for instances where namespaces are created and deleted, when in fact we only use this function to validate the user (if the user is an admin), and use the renderOldAndNewNamespaces function to deal with the specific functionality for CRUD operations. I'm not sure if what I'm trying to say makes sense, and it certainly isn't worthy of holding up the merge of this PR. just a thought.

pkg/webhooks/namespace/namespace.go
```diff
@@ -123,39 +168,40 @@ func (s *NamespaceWebhook) Authorized(request admissionctl.Request) admissionctl
 // Is the request authorized?
 func (s *NamespaceWebhook) authorized(request admissionctl.Request) admissionctl.Response {
 	var ret admissionctl.Response
+
+	// Picking OldObject or Object will suffice for most validation concerns
 	ns, err := s.renderNamespace(request)
```
csheremeta, line 173: considering this is the only place we use `renderNamespace`, I wonder if it would be helpful to combine the logic of `renderNamespace` and `renderOldAndNewNamespaces` and use the resulting function in all places (or if that would, instead, be more confusing?)

pkg/webhooks/namespace/namespace.go
```diff
@@ -123,39 +168,40 @@ func (s *NamespaceWebhook) Authorized(request admissionctl.Request) admissionctl
...
 	if err != nil {
 		log.Error(err, "Couldn't render a Namespace from the incoming request")
 		return admissionctl.Errored(http.StatusBadRequest, err)
 	}
-	// L49-L56
-	// service accounts making requests will include their name in the group
```
csheremeta, line 132: I agree that the comments here with line numbers don't make sense (and are probably just obsolete, I'm guessing from translation to golang), but I think I'd like to see the (attempts at) helpful comments themselves in this line and the one below stay, for readability

pkg/webhooks/namespace/namespace.go
```diff
@@ -123,39 +168,40 @@ func (s *NamespaceWebhook) Authorized(request admissionctl.Request) admissionctl
...
 	for _, group := range request.UserInfo.Groups {
 		if privilegedServiceAccountsRe.Match([]byte(group)) {
 			ret = admissionctl.Allowed("Privileged service accounts may access")
 			ret.UID = request.AdmissionRequest.UID
 			return ret
 		}
 	}
-	// L58-L62
-	// This must be prior to privileged namespace check
```
csheremeta, line 141: this is the second comment I think we could probably leave in

pkg/webhooks/namespace/namespace.go
```diff
@@ -187,12 +232,86 @@ func (s *NamespaceWebhook) authorized(request admissionctl.Request) admissionctl
...
+		}
+		// There were some found
+		return true, fmt.Errorf("Manged OpenShift customers may not directly set certain protected labels (%s) on Namespaces", protectedLabels)
+	} else if req.Operation == v1beta1.Update {
+		// For Updates we must see if the new object is making a change to the old one for any protected labels.
+		// First, let's see if the old object had any protected labels we ought to
+		// care about. If it has, then we can use that resulting list to compare to
+		// the newNamespace for any changes. However, just because the oldNamespace
+		// did not have any protected labels doesn't necessarily mean that we can
+		// ignore potential setting of those label's values in the newNamespace.
```
csheremeta, line 276:
```suggestion
		// ignore potential setting of those labels' values in the newNamespace.
```

pkg/webhooks/namespace/namespace.go
lisa: Good call.

pkg/webhooks/namespace/namespace.go
```diff
@@ -187,12 +232,86 @@ func (s *NamespaceWebhook) authorized(request admissionctl.Request) admissionctl
...
+				return true, fmt.Errorf("Managed OpenShift customers may not change the value or certain protected labels (%s) on Namespaces. %s changed from %s to %s", protectedLabels, labelKey, oldNamespace.Labels[labelKey], newNamespace.ObjectMeta.Labels[labelKey])
+			}
+		}
+	}
+	return false, nil
+}
+
+// doesNamespaceContainProtectedLabels checks the namespace for any instances of
+// protectedLabels and returns a slice of any instances of matches
+func doesNamespaceContainsProtectedLabels(ns *corev1.Namespace) []string {
```
csheremeta, line 305:
```suggestion
func doesNamespaceContainProtectedLabels(ns *corev1.Namespace) []string {
```

pkg/webhooks/namespace/namespace.go
```diff
@@ -187,12 +232,86 @@ func (s *NamespaceWebhook) authorized(request admissionctl.Request) admissionctl
...
+	}
+
+	newNamespace, oldNamespace, err := s.renderOldAndNewNamespaces(req)
+	if err != nil {
+		return true, err
+	}
+	if req.Operation == v1beta1.Create {
+		// For creations, we look to newNamespace and ensure no protectedLabels are set
+		// We don't care about oldNamespace.
+		protectedLabelsFound := doesNamespaceContainsProtectedLabels(newNamespace)
```
csheremeta, line 264:
```suggestion
		protectedLabelsFound := doesNamespaceContainProtectedLabels(newNamespace)
```

pkg/webhooks/namespace/namespace.go
```diff
@@ -187,12 +232,86 @@ func (s *NamespaceWebhook) authorized(request admissionctl.Request) admissionctl
...
+		// First, let's see if the old object had any protected labels we ought to
+		// care about. If it has, then we can use that resulting list to compare to
+		// the newNamespace for any changes. However, just because the oldNamespace
+		// did not have any protected labels doesn't necessarily mean that we can
+		// ignore potential setting of those label's values in the newNamespace.
+
+		// protectedLabelsFoundInOld is a slice of all instances of protectedLabels
+		// that appeared in the oldNamespace that we need to be sure have not
+		// changed.
+		protectedLabelsFoundInOld := doesNamespaceContainsProtectedLabels(oldNamespace)
```
csheremeta, line 281:
```suggestion
		protectedLabelsFoundInOld := doesNamespaceContainProtectedLabels(oldNamespace)
```

pkg/webhooks/namespace/namespace.go
```diff
@@ -187,12 +232,86 @@ func (s *NamespaceWebhook) authorized(request admissionctl.Request) admissionctl
...
+		// protectedLabelsFoundInNew is a slice of all instances of protectedLabels
+		// that appeared in the newNamespace that we need to be sure do not have a
+		// value different than oldNamespace.
+		protectedLabelsFoundInNew := doesNamespaceContainsProtectedLabels(newNamespace)
```
csheremeta, line 285:
```suggestion
		protectedLabelsFoundInNew := doesNamespaceContainProtectedLabels(newNamespace)
```

pkg/webhooks/namespace/namespace.go
lisa: I actually had it combined in a previous (uncommitted) but the logic and utilization ended up being less readable. The method had three returns, and there still had to be logic at this point in the package to know which to use.

I'm certain there is room for more refactoring, but I'm not quite sure it includes combining the logic of both.

csheremeta: nod, I figured the number of returns would make for some difficult reading and logic. this is good (for now :) )

lisa:
/hold

for squashing

lisa:
/unhold

all squashed!

PR #128, opened 2021-01-19
OSD-6264: Prefer upstream CA injection

lisa:
With the deployment and widespread adoption of OSD 4.3, this hack can be
removed, which simplifies this webhook codebase by removing an init
container, extra RBAC, and instead relies on the service-ca-operator
annotation.


lisa:
/hold

for review

lisa:
/cc @rogbas 

@rogbas @jewzaam will not have time to review this and so by random chance you have been co-assigned reviewship with Haoran.

build/selectorsyncset.yaml
```diff
@@ -141,7 +125,7 @@ objects:
       kind: ValidatingWebhookConfiguration
       metadata:
         annotations:
-          managed.openshift.io/inject-cabundle-from: openshift-validation-webhook/webhook-cert
+          service.beta.openshift.io/inject-cabundle: "true"
```
rogbas, line 128: Why are we updating the annotation prefix to `service.beta` ?

build/syncset.go
```diff
@@ -339,12 +315,7 @@ func createValidatingWebhookConfiguration(hook webhooks.Webhook) admissionregv1.
 			Name: fmt.Sprintf("sre-%s", hook.Name()),
 
 			Annotations: map[string]string{
-				"managed.openshift.io/inject-cabundle-from": fmt.Sprintf("%s/webhook-cert", *namespace),
-				// TODO(lseelye): Leave out until Until 4.4 when openshift-ca-operator
-				// supports ValidatingWebhookConfigurations, so as to not fight against
-				// that operator once 4.4 lands.
-				//	// For openshift/service-ca-operator
-				//	"service.beta.openshift.io/inject-cabundle": "true",
+				"service.beta.openshift.io/inject-cabundle": "true",
```
rogbas, line 318: I would add a comment here explaining what the `service.beta.openshift.io/inject-cabundle` annotation will trigger.

build/selectorsyncset.yaml
lisa: @rogbas This is the annotation required by https://github.com/openshift/service-ca-operator

build/syncset.go
lisa: Can do. I fell into the trap of "well the annotation is obvious" but while it does say "inject cert" it is not clear on the face of it what that implies, and what is doing injection. I will update and provide a comment.

build/selectorsyncset.yaml
lisa: In a literal sense the "why" is that this `build/selectorsyncset.yaml` file is generated by the `build/syncset.go` file (`make syncset`), and since the annotations in the `syncset.go` file changed, so too will the generated output.

PR #129, opened 2021-01-20
Use SELinux volume mount flag in syncset make target

lisa:


lisa:
/assign @boranx 
/cc @boranx 

Boran, if the diff I gave you yesterday to test works, can you /lgtm on this please?

boranx (approved):

PR #130, opened 2021-01-20
Add ObjectSelector support to Webhook interface

lisa:
In support of [OSD-3296](https://issues.redhat.com/browse/OSD-3296)

Starting in Kubernetes 1.15, webhooks gained the ability to use a
LabelSelector to "limit which requests are intercepted" based on the
labels of objects that would be sent. This PR adds to the Webhook
interface an `ObjectSelector() *metav1.LabelSelector` method which will
support this. It is recommended to set to `nil` if the particular
webhook has no such ObjectSelector restrictions (so that the resulting
selectorsyncset.yaml file does not include an empty map, for clarity
sake).


boranx:
This lgtm aswell however I'd still wanna get the label from actual maintainers/contributors.

lisa:
@boranx I am the maintainer. :) But I can't /lgtm my own PRs.

PR #136, opened 2021-02-11
OSD-6482: Dedicated-admins should be able to manage CustomDomain

lisa:
This will allow `CustomDomain` objects from `managed.openshift.io` to be
managed by members of the dedicated-admins group. This will also
slightly modify the logic of the `MustGather` section to refer to the
ApiGroup as well as the Kind (for future safety), and switch them to
being references to constants. Further, the check is done outside a loop
so it's a wee bit faster.

### DIFF
--- a/pkg/webhooks/regularuser/regularuser.go
+++ b/pkg/webhooks/regularuser/regularuser.go
@@ -17,14 +17,17 @@ import (
 )
 
 const (
-	WebhookName string = "regular-user-validation"
-	docString   string = `Managed OpenShift customers may not manage any objects in the following APIgroups %s, nor may Managed OpenShift customers alter the ClusterVersion, Node or SubjectPermission objects.`
+	WebhookName       string = "regular-user-validation"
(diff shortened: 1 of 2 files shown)


pkg/webhooks/regularuser/regularuser.go
```diff
@@ -168,18 +171,26 @@ func (s *RegularuserWebhook) authorized(request admissionctl.Request) admissionc
 		ret.UID = request.AdmissionRequest.UID
 		return ret
 	}
+	if utils.SliceContains("dedicated-admins", request.UserInfo.Groups) &&
```
wgordon17, line 174: Does this need to _also_ accounts for `cluster-admins` users?

dustman9000:
/lgtm
thanks Lisa!

pkg/webhooks/regularuser/regularuser.go
lisa: If that's in the spec, yes. Let's chat Tuesday.

PR #137, opened 2021-02-12
WIP: ClusterLogging validation webhook

NautiluX:


lisa:
/hold

pending review and discussion of changes.

pkg/webhooks/clusterlogging/clusterlogging.go
```diff
@@ -0,0 +1,250 @@
...
+	decoder, err := admissionctl.NewDecoder(&s.s)
+	if err != nil {
+		return admissionctl.Errored(http.StatusBadRequest, err)
+	}
+	clusterLogging := &cl.ClusterLogging{}
+	if len(request.OldObject.Raw) > 0 {
+		err = decoder.DecodeRaw(request.OldObject, clusterLogging)
+	} else {
+		err = decoder.DecodeRaw(request.Object, clusterLogging)
+	}
```
lisa, line 146:
```suggestion
// renderClusterLogging decodes an *cl.ClusterLogging from the incoming request.
// If the request includes an OldObject (from an update or deletion), it will be
// preferred, otherwise, the Object will be preferred.
func (s *ClusterloggingWebhook) renderClusterLogging(request admissionctl.Request) (*cl.ClusterLogging, error) {
	decoder, err := admissionctl.NewDecoder(&s.s)
	if err != nil {
		return nil, err
	}
	clusterLogging := &cl.ClusterLogging{}
	if len(request.OldObject.Raw) > 0 {
		err = decoder.DecodeRaw(request.OldObject, clusterLogging)
	} else {
		err = decoder.DecodeRaw(request.Object, clusterLogging)
	}
	if err != nil {
		return nil, err
	}
	return clusterLogging, nil
}

// Authorized implements Webhook interface
func (s *ClusterloggingWebhook) authorized(request admissionctl.Request) admissionctl.Response {
	var ret admissionctl.Response

	clusterLogging, err := s.renderClusterLogging(request)
	if err != nil {
		return admissionctl.Errored(http.StatusBadRequest, err)
	}
```

In other places where objects are decoded, we have the convention to put those operations into their own method ([cf namespace hook](https://github.com/openshift/managed-cluster-validating-webhooks/blob/3031ecc55e118a436db367fd153025d396d01799/pkg/webhooks/namespace/namespace.go#L111-L140)), so I've made a Github code suggestion to do that here. Please have a look and let me know what you think.

go.mod
```diff
@@ -3,12 +3,16 @@ module github.com/openshift/managed-cluster-validating-webhooks
 go 1.14
 
 require (
+	github.com/ViaQ/logerr v1.0.9
```
lisa, line 6:
```suggestion
```

This import is unused.

go.mod
```diff
@@ -3,12 +3,16 @@ module github.com/openshift/managed-cluster-validating-webhooks
...
 	github.com/ghodss/yaml v1.0.1-0.20190212211648-25d852aebe32
-	github.com/golang/mock v1.3.1
+	github.com/golang/mock v1.4.1
 	github.com/openshift/api v3.9.1-0.20191111211345-a27ff30ebf09+incompatible
+	github.com/openshift/cluster-logging-operator v0.0.0-20210209192112-1a2f4a0ea6f3
+	github.com/openshift/elasticsearch-operator v0.0.0-20200722044541-14fae5dcddfd
 	github.com/openshift/hive v1.0.4
-	k8s.io/api v0.18.2
-	k8s.io/apimachinery v0.18.2
+	github.com/pkg/errors v0.9.1
```
lisa, line 13:
```suggestion
```

This import (appears to be?) unused.

pkg/webhooks/clusterlogging/clusterlogging.go
```diff
@@ -0,0 +1,250 @@
...
+
+	if request.Operation == admissionapi.Delete {
+		return admissionctl.Allowed("Allowed to delete ClusterLogging")
+	}
```
lisa, line 150: If we're always allowing deletions, is there a reason to change the rules to match only `CREATE` and `UPDATE` operations? This would save some calls to the webhook and decoding of the object, etc.

pkg/webhooks/clusterlogging/clusterlogging.go
```diff
@@ -0,0 +1,250 @@
...
+
+	if err != nil {
+		log.Error(err, "Couldn't decode ClusterLogging CR from incoming request")
+		return admissionctl.Errored(http.StatusBadRequest, err)
+	}
```
lisa, line 155:
```suggestion
```

If the above suggestion to abstract object decoding to a method is adopted, this short block may be removed.

pkg/webhooks/clusterlogging/clusterlogging.go
```diff
@@ -0,0 +1,250 @@
...
+
+	retentionPolicy := clusterLogging.Spec.LogStore.RetentionPolicy
+
+	appValidator := retentionPolicyValidator{"app", "Set MaxAge to a value <= 7d", esapi.TimeUnit("7d")}
```
lisa, line 159:
```suggestion
	appValidator := retentionPolicyValidator{
		name:       "app",
		hint:       "Set MaxAge to a value <= 7d",
		upperBound: esapi.TimeUnit("7d"),
	}
```

To follow the style elsewhere in the project, please use this longer form struct declaration. The reasoning is that this long form helps readers follow along with the field, and also aids in searching the source file (such as for `hint`, as I did, to understand its purpose).

Alternately, this can be on one line if you deem that to be more readable:

```
	appValidator := retentionPolicyValidator{name: "app", hint: "Set MaxAge to a value <= 7d", upperBound: esapi.TimeUnit("7d")}
```

The auto-GitHub suggestion will apply the multi-line variant.

pkg/webhooks/clusterlogging/clusterlogging.go
```diff
@@ -0,0 +1,250 @@
...
+	ok, ret := appValidator.checkPolicy(retentionPolicy.App)
+	if !ok {
+		return ret
+	}
+
+	infraValidator := retentionPolicyValidator{"infra", "MaxAge must be 0d", esapi.TimeUnit("0d")}
```
lisa, line 165:
```suggestion
	infraValidator := retentionPolicyValidator{
		name:       "infra",
		hint:       "MaxAge must be 0d",
		upperBound: esapi.TimeUnit("0d"),
	}
```

Similar to above.

Single-line suggestion: 
```
	infraValidator := retentionPolicyValidator{name: "infra", hint: "MaxAge must be 0d", upperBound: esapi.TimeUnit("0d")}
```

pkg/webhooks/clusterlogging/clusterlogging.go
```diff
@@ -0,0 +1,250 @@
...
+	ok, ret = infraValidator.checkPolicy(retentionPolicy.Infra)
+	if !ok {
+		return ret
+	}
+
+	auditValidator := retentionPolicyValidator{"audit", "Set MaxAge to a value <= 7d", esapi.TimeUnit("7d")}
```
lisa, line 171:
```suggestion
	auditValidator := retentionPolicyValidator{
		name:       "audit",
		hint:       "Set MaxAge to a value <= 7d",
		upperBound: esapi.TimeUnit("7d"),
	}
```

As above.

```
	auditValidator := retentionPolicyValidator{name: "audit", hint: "Set MaxAge to a value <= 7d", upperBound: esapi.TimeUnit("7d")}
```

pkg/webhooks/clusterlogging/clusterlogging.go
```diff
@@ -0,0 +1,250 @@
...
+	ok, ret = auditValidator.checkPolicy(retentionPolicy.Audit)
+	if !ok {
+		return ret
+	}
+
+	return ret
+}
+
+func le(lhs esapi.TimeUnit, rhs esapi.TimeUnit) (bool, error) {
```
lisa, line 180: Is this a candidate to bring upstream to `esapi` and export for use here?

lisa (changes requested):
Thanks for the PR, @NautiluX. I have made a number of requests for changes, most stemming to unify the style used throughout this repository. Some changes are functional.

Go.mod seems to include things which are unused. Please run `go mod tidy`. I have made (GitHub) suggestions to `go.mod` that remove the unused dependencies. If these changes to go.mod are accepted, `go.sum` will need to be updated as well and recommitted (`go mod tidy`).

Finally, a request which I can't more easily make with GitHub suggestions is around other convention in this repository: We include the request UID in the response for tracking purposes. Your suggested PR does not do this. The other webhooks make this `var ret admissionctl.Response` declaration near the top of their authorizing methods and then follow this:

```go
	ret = admissionctl.Denied("deny reason")
	return ret
```

Please adopt this convention of preserving the request UID.

pkg/webhooks/clusterlogging/clusterlogging.go
NautiluX: They have functions upstream to convert to milliseconds, which would be good enough, but it's not exported. The way the conversion to seconds here is implemented is not accurate for units `>m` but good enough for our purpose. So if the upstream one would be exported, it could be used instead of the one implemented in this PR.

pkg/webhooks/clusterlogging/clusterlogging_test.go
```diff
@@ -0,0 +1,172 @@
...
+	s := fmt.Sprintf(testObjectRaw, appMaxAge, infraMaxAge, auditMaxAge)
+	return s
+}
+
+func Test_InvalidTimeUnit(t *testing.T) {
+	testSuites := []clusterloggingTestSuite{
+		NewTestSuite("7x", "0d", "0d").ExpectNotAllowed(),
+		NewTestSuite("7D", "0d", "0d").ExpectNotAllowed(),
+		NewTestSuite("7S", "0d", "0d").ExpectNotAllowed(),
+		NewTestSuite("m", "0d", "0d").ExpectNotAllowed(),
```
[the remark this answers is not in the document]
NautiluX: Isn't it? :)

pkg/webhooks/clusterlogging/clusterlogging.go
```diff
@@ -0,0 +1,250 @@
+package clusterlogging
+
+import (
+	"fmt"
+	"net/http"
+	"regexp"
+	"strconv"
+
+	cl "github.com/openshift/cluster-logging-operator/pkg/apis/logging/v1"
+	esapi "github.com/openshift/elasticsearch-operator/pkg/apis/logging/v1"
```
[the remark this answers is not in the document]
NautiluX: sounds like a good idea. Will look at it today.

pkg/webhooks/clusterlogging/clusterlogging.go
NautiluX: Done. Thanks for this suggestion!

pkg/webhooks/clusterlogging/clusterlogging.go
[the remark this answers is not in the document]
NautiluX: done

pkg/webhooks/clusterlogging/clusterlogging_test.go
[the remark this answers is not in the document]
lisa: Absolutely. I'm thinking that you've found this repository's next refactor. If you're not careful you'll get volunteered for it!! 😉  😛

pkg/webhooks/clusterlogging/clusterlogging.go
lisa: That's fair. Our team doesn't own that repository so it's not as straightforward to get the feature request in.

lisa:
@NautiluX It appears you'll need a rebase once #141 is updated to address build issues.

NautiluX:
rebased

pkg/webhooks/clusterlogging/clusterlogging.go
```diff
@@ -0,0 +1,267 @@
...
+	} else {
+		err = decoder.DecodeRaw(request.Object, clusterLogging)
+	}
+	if err != nil {
+		return nil, err
+	}
+	return clusterLogging, nil
+}
+
+// Authorized implements Webhook interface
```
lisa, line 153: This comment should be with the exported method. Not sure how I missed that, so my apologies.

NautiluX: yep, good catch

pkg/webhooks/clusterlogging/clusterlogging_test.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+
+	runTests(t, testSuites)
+}
+
+func runTests(t *testing.T, tests []clusterloggingTestSuite) {
+	gvk := metav1.GroupVersionKind{
+		Group:   "quota.openshift.io",
+		Version: "v1",
+		Kind:    "ClusterResourceQuota",
```
boranx, line 125: is this relevant to logging validation hook tests? Why do we have to define this and refer to quota?

NautiluX: copypasta I guess... anyways it's not checked and just needs to be filled for the function to generate a request I think.

NautiluX: done

pkg/webhooks/clusterlogging/clusterlogging_test.go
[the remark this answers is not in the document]
NautiluX: :disappear:

NautiluX:
/hold cancel
discussed with PM, the current configuration of up to 7d for app and 0 for infra and audit are fine.

lisa:
/hold

for reasons mentioned off Github.

lisa (approved):
/lgtm
/approve

NautiluX:
/hold
need to implement a way to only validate ClusterLogging CRs on clusters < 4.7

lisa:
/approve cancel

@NautiluX  With the structural changes I'd like to chat with you about it. I have an idea which I'd like to run past you for consideration. Sadly I did not have time today for a thorough review, but I've seen enough at this point to have some questions and some alternatives to suggest. I'll try and reach you on Slack tomorrow AM before my on-call shift.

pkg/webhooks/utils/utils.go
```diff
@@ -20,6 +21,14 @@ var (
...
 )
 
+func DefaultLabelSelector() metav1.LabelSelector {
+	return metav1.LabelSelector{
+		MatchLabels: map[string]string{
+			"api.openshift.com/managed": "true",
+		},
+	}
+}
+
```
lisa, line 31: Handy!

pkg/webhooks/clusterlogging/clusterlogging.go
```diff
@@ -0,0 +1,285 @@
...
+	if err != nil {
+		return 0, "", fmt.Errorf("unable to parse uint '%s' ", n)
+	}
+	unit := match[2]
+	return number, unit, nil
+}
+
+// SyncSetLabelSelector returns the label selector to use in the SyncSet.
+func (s *ClusterloggingWebhook) SyncSetLabelSelector() metav1.LabelSelector {
+	return metav1.LabelSelector{
```
lisa, line 259: I wonder if there's a way to do a merge from the default with set with the MatxhExpressions.

build/syncset.go
```diff
@@ -436,20 +450,20 @@ func main() {
...
 	}
 	// TODO(lseelye): Until 4.3, place the DaemonSet after the
 	// ValidatingWebhookConfigurations are added to the `encoded` list so that the
 	// initContainer will work (see createDaemonSet and
 	// createValidatingWebhookConfiguration)
-	encoded = append(encoded, runtime.RawExtension{Object: createDaemonSet()})
+	templateResources.Add(metav1.LabelSelector{
+		MatchLabels: map[string]string{
+			"api.openshift.com/managed": "true",
+		},
```
lisa, line 465: Can we move this above the creation of the `createValidatingWebhookConfiguration` and use the `utils.DefaultLabelSelector()` call here?

build/syncset.go
```diff
@@ -388,27 +386,43 @@ func createSelectorSyncSet(resources []runtime.RawExtension) *hivev1.SelectorSyn
...
+}
+
+func (s *SyncSetResourcesByLabelSelector) Add(labelSelector metav1.LabelSelector, object runtime.RawExtension) {
+	for i, entry := range s.entries {
+		if reflect.DeepEqual(entry.labelSelector, labelSelector) {
+			s.entries[i].objects = append(s.entries[i].objects, object)
+			return
+		}
+	}
+	s.entries = append(s.entries, SyncSetResource{labelSelector, []runtime.RawExtension{object}})
```
lisa, line 410: Could you add some comments here to help explain the flow of what's happening, especially on line 406 and 410?

build/syncset.go
```diff
@@ -388,27 +386,43 @@ func createSelectorSyncSet(resources []runtime.RawExtension) *hivev1.SelectorSyn
...
 }
+
+type SyncSetResourcesByLabelSelector struct {
+	entries []SyncSetResource
+}
+
+type SyncSetResource struct {
+	labelSelector metav1.LabelSelector
+	objects       []runtime.RawExtension
+}
```
lisa, line 401: For future us: "Why can't we just use a `map[metav1.LabelSelector]someOtherType`" ??? Maybe a short statement as to what these to structs are for/doing?

lisa (changes requested):
As discussed in the PR review call.

Additionally, I support your idea to move the Syncset mapping logic to a package (utils/syncsetgeneration maybe?).

pkg/webhooks/clusterlogging/clusterlogging.go
lisa: If we don't (because it's not possible), maybe a note here about why?

NautiluX: done

pkg/webhooks/utils/utils.go
NautiluX: thx :)

build/syncset.go
NautiluX: done

build/syncset.go
NautiluX: moved into its own package and refactored the functions a bit so it should be clearer what they do.

build/syncset.go
NautiluX: done

NautiluX:
I thought about returning a `*metav1.LabelSelector` in the interface to better fit the other functions, but it makes the implementation messier than forcing the user to return a default.

lisa:
> I thought about returning a `*metav1.LabelSelector` in the interface to better fit the other functions, but it makes the implementation messier than forcing the user to return a default.

Oh, that is a very good point. We want to force this behaviour, so it's best if it can't be a pointer (read: `nil`).

lisa (approved):

PR #140, opened 2021-02-17
Allow cluster-admins create/update/delete for CustomDomains CR

dustman9000:
This allows users that are part of `cluster-admins` to add/update/delete CustomDomain CRs.


pkg/webhooks/regularuser/regularuser.go
```diff
@@ -171,10 +171,11 @@ func (s *RegularuserWebhook) authorized(request admissionctl.Request) admissionc
 		ret.UID = request.AdmissionRequest.UID
 		return ret
 	}
-	if utils.SliceContains("dedicated-admins", request.UserInfo.Groups) &&
+	if (utils.SliceContains("cluster-admins", request.UserInfo.Groups) ||
+		utils.SliceContains("dedicated-admins", request.UserInfo.Groups)) &&
 		request.Kind.Kind == customDomainKind &&
 		request.Kind.Group == customDomainGroup {
```
lisa, line 177: @dustman9000 I think it's time to wrap this logic in a function call. I was going to make a (GitHub) suggestion to put the slice checks on one line, and then noticed the trailing `&&`. Four lines of logic begs for a purpose-built function. Please add one with a sensible name.

lisa (changes requested):

lisa (approved):
/lgtm
/approve

PR #141, opened 2021-02-17
fix: dockerfile image

boranx:
registry.ci is the new image registry that replaces (old)registry.svc as per https://docs.ci.openshift.org/docs/how-tos/use-registries-in-build-farm/. We can transition to use the new as the public images are supported and available by CI.


fahlmant (approved):
/lgtm

lisa:
/lgtm
/approve

@fahlmant Please /unhold when you're ready for this to merge.

PR #146, opened 2021-03-09
Allow backplane-cluster-admin user to operate as an SRE

cblecker:
further to OSD-6373, this permits the backplane-cluster-admin user to operate as if it is an SRE

/cc @lisa @sedroche


pkg/webhooks/regularuser/regularuser.go
```diff
@@ -181,6 +182,13 @@ func (s *RegularuserWebhook) authorized(request admissionctl.Request) admissionc
 		ret.UID = request.AdmissionRequest.UID
 		return ret
 	}
+	for _, user := range adminUsers {
+		if request.AdmissionRequest.UserInfo.Username == user {
```
lisa, line 186: `if utils.SliceContains(request.AdmissionRequest.UserInfo.Username, adminUsers` suitable here?

cblecker: oh, yeah it would :)

PR #147, opened 2021-03-10
Add backplane-cluster-admin

sedroche:
Jira; https://issues.redhat.com/browse/OSD-6373

backplane has added a new cluster-admin user for elevating privileges.

I encountered this today:
```
$ oc adm must-gather --as backplane-cluster-admin
[must-gather      ] OUT Using must-gather plugin-in image: quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:b52c1d685b2ec2c674d8abb6ed0e69fd5e71f8c4e0c086fd6d2ca0e3f498470f
Error from server (Prevented from accessing Red Hat managed namespaces. Customer workloads should be placed in customer namespaces, and should not match this regular expression: (^kube.*|^openshift.*|^default$|^redhat.*)): admission webhook "namespace-validation.managed.openshift.io" denied the request: Prevented from accessing Red Hat managed namespaces. Customer workloads should be placed in customer namespaces, and should not match this regular expression: (^kube.*|^openshift.*|^default$|^redhat.*)
```
This PR should fix

PR #148, opened 2021-03-10
Add backplane-cluster-admin as admin user

wanghaoran1988:
This is a follow up to https://github.com/openshift/managed-cluster-validating-webhooks/pull/147


wanghaoran1988:
cc @sedroche @lisa

csheremeta:
this lgtm, but will unit tests be added for these?

wanghaoran1988:
@csheremeta The existing tests already covered that, we have tests for privileged user "system:admin", and I just add the new user in that slice, I don't think we need test all the users in the "admin user" group, WDYT?

csheremeta:
I'm totally biased towards overzealous testing. @lisa what are your thoughts here?

wanghaoran1988:
@csheremeta Sounds good, tests added.

PR #149, opened 2021-03-10
Update retention policy boundaries

NautiluX:
* 1h <= app <= 7d
* infra == 1h
* audit == 1h

jira: [OSD-6701](https://issues.redhat.com/browse/OSD-6701)


RiRa12621:
we have requested the docs to be updated.
imho would shouldn't be waiting for it.
setting the retention to 0 has caused one customer incident and we should adjust this rather quickly 
will still leave it up to you to remove the hold @lisa 

/lgtm

lisa:
See also https://github.com/openshift/openshift-docs/issues/30300

PR #150, opened 2021-03-23
Allow logging installs (continue to block upgrades)

csheremeta:


pkg/webhooks/subscription/subscription.go
```diff
@@ -130,10 +130,17 @@ func (s *SubscriptionWebhook) authorized(request admissionctl.Request) admission
 		return ret
 	}
 
+	// since OSDOCS-1784 is done, we are now allowing installs - March 23, 2021
+	if request.Operation == v1beta1.Create {
+		ret = admissionctl.Allowed("Base decisions for logging installation on RBAC")
+		ret.UID = request.AdmissionRequest.UID
+		return ret
+	}
+
```
lisa, line 139: With the change on line 35 to remove `CREATE` as a verb this webhook will match against, this block does nothing?

lisa: Not off the top of my head, but if we're not matching `CREATE`, all of those unit tests using `CREATE` can be removed; they're not testing functionality of this webhook.

csheremeta: totally fair, but I did want to make sure that those users could actually `CREATE` the subscriptions. the failures from the unit tests indicated that the webhook was still catching those requests. I obviously didn't test this e2e - is this just how the unit tests work, even if the webhook itself shouldn't catch/interfere with those requests?

lisa: One thing to keep in mind with the unit tests is that they are simulating shoving an `AdmissionRequest` into the code handler without any real Kubernetes control plane to do the filtering. That means that you can send a `DELETE` verb to a webhook that doesn't care about deletions at all, and some thing will happen. Perhaps sending a request with a verb that the hook has no rules should be considered "undefined behaviour(?)"

If the webhook does not intercept `CREATE` then the native Kubernetes RBAC rules will handle it, and this webhook doesn't get involved.

csheremeta: alright, that makes sense. I can rip all those unit tests out. we should probably try this out in stage before promotion, in that case?

lisa: Yes, better to be safe than sorry, especially letting e2e exercise these.

csheremeta: 👍🏻

lisa:
/lgtm
/approve

`+37 −730` 😍

lisa:
Thanks bot! @csheremeta need also `make syncset` committed. :)

PR #158, opened 2021-05-26
OSD-7102: move to k8s.io/api/admission/v1

lisa:
A number of things have changed to support this change with the most obvious being the transition from `k8s.io/api/admission/v1beta1` to `k8s.io/api/admission/v1`. This has required the upgrading of _many_ packages (listed in `go.mod`):

* k8s.io/api to v0.21.1
* k8s.io/apimachinery to v0.21.1
* k8s.io/client-go to v0.21.1
* sigs.k8s.io/controller-runtime to v0.8.3
* github.com/openshift/api to `e273a339932a`

Other minor changes as well to go.mod in support of those package version bumps.

Notably is the addition of klog v2.9.0 because the logging facilities from controller-runtime have changed. This unfortunately means the log formatting has changed as well.

Secondary is the openshift/api package (used for syncset generation), which has deeper requirements that were conflicting. It depends on the k8s.io v0.21.1 version, so that is why it's used here, too.

All places in code have been updated to reflect the change to use the `v1` API.

For development purposes I have also modified the Makefile to allow for people to specify a `V=1` to `make` to see commands that are being run. The default behaviour is still to hide this and act as before; sample usage: `make V=1 test`.


lisa:
Running osde2e tests locally have a number of failures that seem to relate here. I'm not sure why, but definitely don't unhold this.

pkg/helpers/response.go
```diff
@@ -5,20 +5,21 @@ import (
...
 
 // SendResponse Send the AdmissionReview.
 func SendResponse(w io.Writer, resp admissionctl.Response) {
-
 	encoder := json.NewEncoder(w)
 	responseAdmissionReview := admissionapi.AdmissionReview{
 		Response: &resp.AdmissionResponse,
 	}
+	responseAdmissionReview.APIVersion = "admission.k8s.io/v1"
+	responseAdmissionReview.Kind = "AdmissionReview"
```
cblecker, line 22:
```suggestion
	responseAdmissionReview := admissionapi.AdmissionReview{
		APIVersion: admissionapi.SchemeGroupVersion.String(),
		Kind: "AdmissionReview",
		Response: &resp.AdmissionResponse,
	}
```

cblecker (commented):
One comment, otherwise seems good to me

cblecker:
oh, looks like it needs a gofmt

NautiluX (commented):
/lgtm

PR #159, opened 2021-06-10
Authorize NetNamespace In Regular User Webhook

dustman9000:
This adds authorization for [NetNamespace](https://docs.openshift.com/container-platform/4.5/rest_api/network_apis/netnamespace-network-openshift-io-v1.html).

This also fixes up some test utils to allow for additional parameters such as Name and Namespace.

https://github.com/openshift/managed-cluster-config/pull/839 depends on this being merged first, such that dedicated-admins cannot update protected namespaces.

https://issues.redhat.com/browse/OSD-7367

### DIFF
--- a/pkg/webhooks/regularuser/regularuser.go
+++ b/pkg/webhooks/regularuser/regularuser.go
@@ -3,8 +3,9 @@ package regularuser
 import (
 	"fmt"
 	"strings"
-	"sync"
 
+	networkv1 "github.com/openshift/api/network/v1"
+	"github.com/openshift/managed-cluster-validating-webhooks/pkg/webhooks/namespace"
 	"github.com/openshift/managed-cluster-validating-webhooks/pkg/webhooks/utils"
 	admissionv1 "k8s.io/api/admission/v1"
 	admissionregv1 "k8s.io/api/admissionregistration/v1"
@@ -23,6 +24,8 @@ const (
 	mustGatherGroup   string = "managed.openshift.io"
 	customDomainKind  string = "CustomDomain"
 	customDomainGroup string = "managed.openshift.io"
+	netNamespaceKind  string = "NetNamespace"
+	netNamespaceGroup string = "network.openshift.io"
 )
 
 var (
@@ -88,14 +91,22 @@ var (
 				Scope:       &scope,
 			},
 		},
+		{
+			Operations: []admissionregv1.OperationType{"*"},
+			Rule: admissionregv1.Rule{
+				APIGroups:   []string{"network.openshift.io"},
+				APIVersions: []string{"*"},
+				Resources:   []string{"netnamespaces", "netnamespaces/*"},
+				Scope:       &scope,
+			},
+		},
 	}
 	log = logf.Log.WithName(WebhookName)
 )
 
 // RegularuserWebhook protects various objects from unauthorized manipulation
 type RegularuserWebhook struct {
-	mu sync.Mutex
-	s  runtime.Scheme
+	s runtime.Scheme
 }
 
 func (s *RegularuserWebhook) Doc() string {
@@ -191,6 +202,11 @@ func (s *RegularuserWebhook) authorized(request admissionctl.Request) admissionc
 		ret.UID = request.AdmissionRequest.UID
 		return ret
 	}
+	if isNetNamespaceAuthorized(s, request) {
+		ret = admissionctl.Allowed("Management of NetNamespace CR is authorized")
+		ret.UID = request.AdmissionRequest.UID
+		return ret
+	}
 	if utils.SliceContains(request.AdmissionRequest.UserInfo.Username, adminUsers) {
 		ret = admissionctl.Allowed("Specified admin users are allowed")
 		ret.UID = request.AdmissionRequest.UID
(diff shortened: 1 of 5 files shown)


pkg/testutils/testutils.go
```diff
@@ -66,6 +71,17 @@ func CreateFakeRequestJSON(uid string,
 			},
 		},
 	}
+	// Allow for additional optional parameters
```
lisa, line 74: What are these for?

pkg/webhooks/regularuser/regularuser.go
```diff
@@ -225,6 +231,18 @@ func isCustomDomainAuthorized(request admissionctl.Request) bool {
...
+// isNetNamespaceAuthorized check if request is authorized for NetNamespace CR
+func isNetNamespaceAuthorized(request admissionctl.Request) bool {
+	fmt.Printf("%s\n", request.Name)
+	return ((utils.SliceContains("cluster-admins", request.UserInfo.Groups) ||
+		utils.SliceContains("dedicated-admins", request.UserInfo.Groups)) &&
+		request.Kind.Kind == netNamespaceKind &&
+		request.Kind.Group == netNamespaceGroup &&
+		request.Name != "" &&
+		!namespace.PrivilegedNamespaceRe.Match([]byte(request.Name)) &&
+		!namespace.BadNamespaceRe.Match([]byte(request.Name)))
```
lisa, line 243: Can we tidy this, please?

pkg/webhooks/regularuser/regularuser.go
```diff
@@ -225,6 +231,18 @@ func isCustomDomainAuthorized(request admissionctl.Request) bool {
 		request.Kind.Group == customDomainGroup)
 }
 
+// isNetNamespaceAuthorized check if request is authorized for NetNamespace CR
+func isNetNamespaceAuthorized(request admissionctl.Request) bool {
+	fmt.Printf("%s\n", request.Name)
```
lisa, line 236: We use `log`, but this seems leftover for development; let's kick it out.

lisa (changes requested):
I think you also need to update the rules for regular user hook.

Also a couple questions and other feedback.

/hold

pkg/webhooks/regularuser/regularuser.go
dustman9000: Yeah, this was leftover from debugging. Will remove.

pkg/testutils/testutils.go
dustman9000: I've added an optional parameter (`opts`) to allow for a test to pass in Name and/or Namespace to the request. For NetNamespaces, we are checking if it's a valid Name.

pkg/webhooks/regularuser/regularuser.go
dustman9000: Sure, I've added a helper function `isNetNamespaceValid()`

dustman9000:
> I think you also need to update the rules for regular user hook.
> 
> Also a couple questions and other feedback.
> 
> /hold

Thanks, I think I've addressed your comments.  I've also added the rule for `netnamespaes.network.openshift.io`.

pkg/testutils/testutils.go
lisa: I gather that from the code, but what issue is being addressed by the change? How are they used by the API server with an incoming payload, and why is it different than the payload in the `object`?

pkg/webhooks/regularuser/regularuser.go
lisa: What I'm looking for is to break these so they're not compound ifs with ors, ands, and equality checks in. It's hard to read and parse.

Consider, for example an improvement:

```go
if request.Name == "" {
	return false
}
// some other check
```

Even better would be to encapsulate the semantic meaning behind what it means for `request.Name` to be equal to an empty string. As a human, I can't understand _why_ it's needed, but if it was encapsulated in a function, with comments around it, it will help future contributors understand.

We can trust the compiler to optimize the logic statements so the computers can run them, let's focus on optimizing for humans to read them.

pkg/testutils/testutils.go
dustman9000: The problem is that the current test utils do not allow for tests to set the Name and/or Namespace in the request. As defined in the [AdmissionRequest](https://github.com/kubernetes/api/blob/master/admission/v1beta1/types.go#L92):
```
        // Name is the name of the object as presented in the request.  On a CREATE operation, the client may omit name and
	// rely on the server to generate the name.  If that is the case, this field will contain an empty string.
	// +optional
	Name string `json:"name,omitempty" protobuf:"bytes,5,opt,name=name"`
```
Adding optional parameters to this function allows tests to both set the Name and Request.Name.
The webhook auth code is checking against the Request.Name, not the object or oldObject name.

pkg/webhooks/regularuser/regularuser.go
dustman9000: The ask is a little out of scope and not consistent with the other functions, but I went ahead and changed `isNetNamespaceValid` to be a little more readable.

pkg/testutils/testutils.go
dustman9000: I've updated the webhook and test code to check the `Request.Object.Name` as opposed to checking the `Request.Name` to limit the amount of churn in the testutils code.

lisa: Can you help me understand why we can't rely on the contents in the incoming`AdmissionRequest`'s `Object` (or `OldObject`, verb depending)? I do not want to complicate things more unless it's a requirement.

It would help if you could post the incoming request into this webhook for each verb (perhaps obfuscating the `userInfo` section).

dustman9000: As I've said above, I've changed the code to rely on the AdmissionRequest Object, not the Name as to not introduce complexity to the code. Please see the updated code. Thanks.

pkg/webhooks/regularuser/regularuser.go
```diff
@@ -225,6 +241,43 @@ func isCustomDomainAuthorized(request admissionctl.Request) bool {
...
+	}
+	netNamespace := &networkv1.NetNamespace{}
+	if len(request.Object.Raw) == 0 {
+		return false
+	} else {
+		err := decoder.Decode(request, netNamespace)
+		if err != nil {
+			return false
+		}
+	}
```
lisa, line 268:
```suggestion
	}
	err = decoder.Decode(request, netNamespace)
	if err != nil {
		return false
	}
```

From linter:

> if block ends with a return statement, so drop this else and outdent its block

pkg/webhooks/regularuser/regularuser.go
```diff
@@ -225,6 +241,43 @@ func isCustomDomainAuthorized(request admissionctl.Request) bool {
...
+	// Check if name is empty
+	if len(netNamespace.Name) == 0 {
+		return false
+	}
```
lisa, line 272: This is not a very useful comment because we can read the following line and understand that it's a check to see if the length of the name is 0. A better comment would be to explain why this is important.

Additionally, an object's `Name` [is guaranteed to be non-empty](https://github.com/kubernetes/community/blob/b345baf47f2b8a55acf9344f0f0412f49b2a3179/contributors/design-proposals/architecture/identifiers.md), so this check can also be removed.
```suggestion

```

lisa (changes requested):
Added feedback/suggestions.

Also note that you'll need to add the networkv1 scheme in the `NewWebhook()` function.

pkg/webhooks/regularuser/regularuser.go
dustman9000: Done.

pkg/webhooks/regularuser/regularuser.go
dustman9000: This is now fixed, but you should probably fix the rest of the code base as well. For example, right [here](https://github.com/openshift/managed-cluster-validating-webhooks/blob/master/pkg/webhooks/namespace/namespace.go#L134).

lisa: Definitely.

lisa (approved):
/lgtm
/approve

Good first pass at this. Thank you for the contribution.

dustman9000:
This is ready to merge, thanks.
/hold cancel

PR #160, opened 2021-08-02
[OSD-7327] Apply owner label to resp.AuditAnnotations

dkeohane:
[Ref OSD-7327](https://issues.redhat.com/browse/OSD-7327)

- In an attempt to narrow the scope of ValidatingWebhookConfigurations Splunk alerts, I'm adding an owner label which can be filtered on in Splunk alerts. This is being done to avoid raising alerts to SREP for customer owned ValidatingWebhookConfigurations


georgettica:
/approve
looks good to me but maybe a maintainer can actually do it

PR #162, opened 2021-08-19
Update golang images to 1.16

cblecker:
This PR updates the golang images for managed-cluster-validating-webhooks to go 1.16. Additionally, it modifies the Makefile to copy [this stanza](https://github.com/openshift/boilerplate/blob/328a6a5330ca758439829122dbdfafca9dcc79cf/boilerplate/test/test-base-convention/standard.mk#L31-L38) from boilerplate make everything compatible with the openshift/release image, and the Dockerfile is slightly modified to download and cache the modules in a layer in order to improve build speed during development.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,17 +13,21 @@ IMG_REGISTRY ?= quay.io
 IMG_ORG ?= app-sre
 IMG ?= $(IMG_REGISTRY)/$(IMG_ORG)/${BASE_IMG}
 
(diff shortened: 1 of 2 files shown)


lisa:
/lgtm
/approve

Thanks for this PR Christoph. It's nice to unify the images here. Maybe next iteration we can use the syncset image as a build arg for the Dockerfile. ;-)

PR #166, opened 2021-09-24
OSD-8075 - Utilize namespaces from managed-namespaces configmap

tnierman:
Completes: https://issues.redhat.com/browse/OSD-8075

---
These changes update managed-cluster-validating-webhooks to utilize the new managed-namespaces configmap when authorizing requests. 

The configmap is read once when the webhooks are first deployed, and the list of namespaces is held in memory for comparison, minimizing any latency issues that might be caused when requesting the ConfigMap from the apiserver.


lisa:
/hold

for review

dofinn:
@lisa  is this still under review?

lisa:
> @lisa is this still under review?

I have not heard anything from James, so I would say so.

jharrington22:
@dustman9000 @tnierman provided the changes here are reading the ConfigMap once during initialization it mitigates the concern of availability and any additional latency added to webhooks. Downsides are we'll have to manage ensuring the ConfigMap in memory is in line with what is configured in MCC however we will address this in future. We are creating a task to change this architecture to watch the ConfigMap out of band and refresh the in memory list, for the time being the list doesn't change very often so it should be easily managed.

@lisa it'd be great if you could add any further thoughts here. 

@tnierman could you confirm that this is ready to go?

tnierman:
@jharrington22 We'll need to update this PR to include the new `ocp-namespaces` and `addons-namespaces` configMaps. Once that's done, I think this PR should be ready.

tnierman:
After discussing, it was deemed preferable to generate the namespace lists as hard-coded values, rather than dynamically reading in from a ConfigMap. 

Closing this PR in favor of #168

PR #168, opened 2021-11-19
OSD-8075 - Enforce webhooks via generated namespace lists.

tnierman:
Completes: https://issues.redhat.com/browse/OSD-8075

---

Follow-up to https://github.com/openshift/managed-cluster-validating-webhooks/pull/166

Adds a `Makefile` rule to generate `PrivilegedNamespaces` and `PrivilegedServiceAccounts` lists from the relevant configMaps in [managed-cluster-config](https://github.com/openshift/managed-cluster-config/tree/master/deploy/osd-managed-resources).

Webhooks will be enforced using these lists, rather than using regular expressions covering all `openshift-.*|kube-.*|redhat-.*` namespaces & serviceaccounts.


pkg/config/namespaces.go
```diff
@@ -0,0 +1,186 @@
...
+package config
+
+var ConfigMapSources = []string{
+	"openshift-monitoring/addons-namespaces",
+	"openshift-monitoring/managed-namespaces",
+	"openshift-monitoring/ocp-namespaces",
+}
+
+var PrivilegedNamespaces = []string{
+	"^redhat.*",
```
dustman9000, line 12: Do we need the regex anchors at all anymore since this are just fixed strings now?

pkg/webhooks/utils/utils.go
```diff
@@ -38,6 +39,16 @@ func SliceContains(needle string, haystack []string) bool {
 	return false
 }
 
+func RegexSliceContains(needle string, haystack []string) bool {
```
dustman9000, line 42: Is the Regex required anymore if the list is just fixed strings? Changing to a simple string lookup could eek out a bit more performance.

pkg/config/namespaces.go
tnierman: The regex is a holdover from [the current implementation](https://github.com/openshift/managed-cluster-validating-webhooks/blob/master/pkg/webhooks/namespace/namespace.go#L22). 

I don't have any qualms about removing the regex per se, but my understanding was that we'd like to keep the current behavior with regard to namespaces that are prefixed with `redhat`. I believe this is because the current namespace lists aren't comprehensive enough to cover all layered product namespaces, which still need protection from webhooks.

I can't think of any way to accomplish both, but if anyone has suggestions, I'm certainly open to improvements

pkg/webhooks/utils/utils.go
tnierman: I think this is blocked by the above comment. Definitely open to removing this function if we can figure out a way to unblock

pkg/config/namespaces.go
dustman9000: Ah that makes sense, I didn't see the redhat wildcard regex.

pkg/webhooks/utils/utils.go
dustman9000: Yes, disregard as I didn't see the `redhat-*` wildcard regex.

README.md
```diff
@@ -27,6 +27,10 @@ A framework supporting validating webhooks for OpenShift.
 
 Ensure the git branch is current and run `make syncset`. The updated Template will be  [build/selectorsyncset.yaml](build/selectorsyncset.yaml) by default.
 
+## Updating PrivilegedNamespaces and PrivilegedServiceAccounts
```
lisa, line 30: Please update the table of contents.

pkg/config/generate/namespaces.go
```diff
@@ -0,0 +1,162 @@
+// +build ignore
```
lisa, line 1:
```diff
lseelye@rhodium managed-cluster-validating-webhooks $ make vet
/Users/lseelye/git/github.com/openshift/managed-cluster-validating-webhooks/pkg/config/generate/namespaces.go
diff -u /Users/lseelye/git/github.com/openshift/managed-cluster-validating-webhooks/pkg/config/generate/namespaces.go.orig /Users/lseelye/git/github.com/openshift/managed-cluster-validating-webhooks/pkg/config/generate/namespaces.go
--- /Users/lseelye/git/github.com/openshift/managed-cluster-validating-webhooks/pkg/config/generate/namespaces.go.orig	2021-11-25 10:22:32.000000000 -0500
+++ /Users/lseelye/git/github.com/openshift/managed-cluster-validating-webhooks/pkg/config/generate/namespaces.go	2021-11-25 10:22:32.000000000 -0500
@@ -1,3 +1,4 @@
+//go:build ignore
 // +build ignore

 package main
make: *** [vet] Error 1
```

pkg/config/generate/namespaces.go
```diff
@@ -0,0 +1,162 @@
...
+)
+
+var namespaceFiles = []string{
+	"addons-namespaces.ConfigMap.yaml",
+	"managed-namespaces.ConfigMap.yaml",
+	"ocp-namespaces.ConfigMap.yaml",
+}
+
+var (
+	// Base lists - default values which will always be enforced regardless of MCC
```
lisa, line 25:
```suggestion
	// Base lists - default values which will always be enforced regardless of managed-cluster-config
```

de-acronymize

pkg/webhooks/regularuser/regularuser.go
```diff
@@ -26,12 +28,15 @@ const (
 	customDomainGroup string = "managed.openshift.io"
 	netNamespaceKind  string = "NetNamespace"
 	netNamespaceGroup string = "network.openshift.io"
+	// Enforce for the following namespaces, regardless of what's configured in MCC
```
lisa, line 31:
```suggestion
	// Enforce for the following namespaces, regardless of what's configured in managed-cluster-config
```

Why isn't this `netNamespaceNS` also in `hookconfig`?

lisa (changes requested):

pkg/webhooks/regularuser/regularuser.go
tnierman: My thinking was that namespaces listed in `hookconfig` are global and unknown the webhooks. The `netNamespaceNS` namespaces seem necessary in all cases (regardless of what `managed-cluster-config` says) to enforce proper net-namespace behavior based on [the existing unit tests](https://github.com/openshift/managed-cluster-validating-webhooks/blob/master/pkg/webhooks/regularuser/regularuser_test.go#L586-L620), so I hard-coded them into this specific hook to ensure that behavior is preserved.

pkg/webhooks/regularuser/regularuser.go
```diff
@@ -266,8 +271,9 @@ func isNetNamespaceValid(s *RegularuserWebhook, request admissionctl.Request) bo
 		return false
 	}
 	// Check if the name is bad or is privileged
-	if namespace.PrivilegedNamespaceRe.Match([]byte(netNamespace.Name)) ||
-		namespace.BadNamespaceRe.Match([]byte(netNamespace.Name)) {
+	if netNamespaceRE.Match([]byte(netNamespace.Name)) ||
+		namespace.BadNamespaceRe.Match([]byte(netNamespace.Name)) ||
+		hookconfig.IsPrivilegedNamespace(netNamespace.Name) {
```
lisa, line 276: Are net namespaces privileged? It's strange to have the comment say check if X or Y and then have three things being checked. These can be "annotated" in this way, it seems:

```go
	// Check if the name is bad or is privileged
	if netNamespaceRE.Match([]byte(netNamespace.Name)) || // privileged
		namespace.BadNamespaceRe.Match([]byte(netNamespace.Name)) || // "bad" (domain names suffixes)
		hookconfig.IsPrivilegedNamespace(netNamespace.Name) { // privileged
		return false
	}
```

This is as if the comment reads: Check if X or Y or X. I'd like to see the idea of privileged namespaces consolidated into the one single function since if it's privileged _here_, it should be considered privileged _everywhere_ without needing this X or Y or X construct.

lisa:
Looks like something has changed and now auth is required to pull the golang image? Refer to https://ci.ext.devshift.net/job/openshift-managed-cluster-validating-webhooks-gh-pr-check/429/ and maybe check with the folks responsible for the CI environment to see what changed and how to accommodate without breaking local dev.

pkg/webhooks/regularuser/regularuser.go
tnierman: Removed the hard-coded namespaces from the `regularuser` webhook after discussing with @dustman9000, who implemented the net-namespaces enforcement logic for the hook.

pkg/webhooks/regularuser/regularuser.go
tnierman: Now that this conditional is back to a single `OR`, do you think it's still necessary to add the annotation to each line?

lisa (approved):
/lgtm

NautiluX (commented):
there is an outdated comment it seems, but other than that
/lgtm
unhold when you're ready to merge
/hold

NautiluX:
oh I'm not an approver, so you need another review anyways :D
/unhold

robotmaxtron (approved):

PR #169, opened 2021-11-30
add validation webhook for scc

bmeng:
fixes [OSD-8315](https://issues.redhat.com/browse/OSD-8315)

### DIFF
--- a/pkg/webhooks/scc/scc.go
+++ b/pkg/webhooks/scc/scc.go
@@ -21,13 +21,12 @@ const (
 )
 
 var (
-	timeout        int32 = 2
-	log                  = logf.Log.WithName(WebhookName)
-	anyuidPriority int32 = 10
-	scope                = admissionregv1.ClusterScope
-	rules                = []admissionregv1.RuleWithOperations{
+	timeout int32 = 2
+	log           = logf.Log.WithName(WebhookName)
+	scope         = admissionregv1.ClusterScope
+	rules         = []admissionregv1.RuleWithOperations{
 		{
-			Operations: []admissionregv1.OperationType{"CREATE", "UPDATE", "DELETE"},
+			Operations: []admissionregv1.OperationType{"UPDATE", "DELETE"},
 			Rule: admissionregv1.Rule{
 				APIGroups:   []string{"security.openshift.io"},
 				APIVersions: []string{"*"},
@@ -72,7 +71,7 @@ func (s *SCCWebHook) Authorized(request admissionctl.Request) admissionctl.Respo
 func (s *SCCWebHook) authorized(request admissionctl.Request) admissionctl.Response {
 	var ret admissionctl.Response
 
-	oldScc, newScc, err := s.renderSCC(request)
+	scc, err := s.renderSCC(request)
 	if err != nil {
 		log.Error(err, "Couldn't render a SCC from the incoming request")
 		return admissionctl.Errored(http.StatusBadRequest, err)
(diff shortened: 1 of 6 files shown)


bmeng:
@NautiluX could you help review the webhook? the test failed due to the image pull.

bmeng:
@csheremeta  @rogbas Could you help review?

pkg/webhooks/scc/scc.go
```diff
@@ -0,0 +1,208 @@
...
+		}
+	}
+	return false
+}
+
+// SCCwithHigherPriority checks if the created SCC has the higher priority
+// than 10 (default to anyuid)
+func SCCwithHigherPriority(scc *securityv1.SecurityContextConstraints) bool {
+	if scc.Priority != nil {
+		if *scc.Priority > anyuidPriority {
```
NautiluX, line 143: can you elaborate why it's needed to prevent customers from creating SCCs with higher priority than anyuid? I think it's not described in the card. I assume it is so the default used if none is specified doesn't change, but I'm not 100% sure?

pkg/webhooks/scc/scc.go
```diff
@@ -0,0 +1,208 @@
...
+
+// renderSCC render the SCC object from the requests
+func (s *SCCWebHook) renderSCC(request admissionctl.Request) (*securityv1.SecurityContextConstraints, error) {
+	decoder, err := admissionctl.NewDecoder(&s.s)
+	if err != nil {
+		return nil, err
+	}
+	scc := &securityv1.SecurityContextConstraints{}
+	if len(request.OldObject.Raw) > 0 {
+		err = decoder.DecodeRaw(request.OldObject, scc)
```
NautiluX, line 118: if a SCC is created with a priority of 1, and then a request is sent to update it to 20, will OldObject.Raw contain the 1 or 20?

pkg/webhooks/scc/scc.go
```diff
@@ -0,0 +1,208 @@
...
+	if request.Operation == admissionv1.Update {
+		if isDefaultSCC(scc) {
+			ret = admissionctl.Denied("Modifying default SCCs is not allowed")
+			ret.UID = request.AdmissionRequest.UID
+			return ret
+		}
+	}
+
+	if request.Operation == admissionv1.Create {
+		if SCCwithHigherPriority(scc) {
```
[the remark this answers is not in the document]
NautiluX: why do we need this check only on create?

NautiluX (changes requested):

pkg/webhooks/scc/scc.go
bmeng: Yeah, that a good point, if we are not going to allow the creation of higher priority, we should also limit the updating as well.

pkg/webhooks/scc/scc.go
bmeng: We were seeing problem due to the customer created higher priority SCC and the `anyuid` cannot be applied to pods . see OHSS-7388
I am assuming this might be a potential issue which we can restrict via webhook. WDYT?

pkg/webhooks/scc/scc.go
```diff
@@ -0,0 +1,208 @@
...
+
+	if request.Operation == admissionv1.Update {
+		if isDefaultSCC(scc) {
+			ret = admissionctl.Denied("Modifying default SCCs is not allowed")
+			ret.UID = request.AdmissionRequest.UID
+			return ret
+		}
+	}
+
+	if request.Operation == admissionv1.Create {
```
lisa, line 97: `switch` statement would be better here.

pkg/webhooks/scc/scc.go
NautiluX: I see the latest commits address updates, but I'm still not sure if OldObject is the right choice for updating resources.

bmeng: Hi Manuel,
I update the `renderSCC` function to read both the old and new object.

So the validation will be:
```
when operation = Create:
    validate the newObject
when operation = Delete:
    validate the oldObject
when operation = Update:
    validate both newObject and oldObject
```

NautiluX: cool. Would it be possible to cover this behavior with a test case? Did you verify the behavior on update? Is the old object the existing one and the new object the one sent to OpenShift?
Do we actually need to validate the old object if the new object is there?

pkg/webhooks/scc/scc.go
```diff
@@ -0,0 +1,189 @@
...
+// GetURI implements Webhook interface
+func (s *SCCWebHook) GetURI() string {
+	return "/" + WebhookName
+}
+
+// Validate implements Webhook interface
+func (s *SCCWebHook) Validate(request admissionctl.Request) bool {
+	valid := true
+	valid = valid && (request.UserInfo.Username != "")
+	valid = valid && (request.Kind.Kind == "SecurityContextConstraints")
```
bmeng, line 140: @NautiluX I found the problem that why it does not work before. The resource scc has some different definition with other. The Kind here, generally the resources will use the singular, but it is plural here...

I have updated the webhook based on our discussion and tested on my staging cluster. Please help take a look.

NautiluX: cool glad local testing is working now as well :)

pkg/webhooks/scc/scc.go
```diff
@@ -0,0 +1,189 @@
...
+			return ret
+		}
+	case admissionv1.Update:
+		if isDefaultSCC(scc) {
+			log.Info(fmt.Sprintf("Updating operation detected on default SCC: %v", scc.Name))
+			ret = admissionctl.Denied(fmt.Sprintf("Modifying default SCCs %v is not allowed", defaultSCCs))
+			ret.UID = request.AdmissionRequest.UID
+			return ret
+		}
+	}
```
NautiluX, line 95: I wonder if you could simplify this block somehow, since for all the cases it seems there is 90% in common.

NautiluX: i.e. aren't the Delete and Update operations filtered already in line 29?
Couldn't we just deny all of the requests if they access a DefaultSCC then?

NautiluX (changes requested):

pkg/webhooks/scc/scc.go
bmeng: But we still need to print the different logs to when the different operations being triggered?

NautiluX: yes. As discussed on Slack, I'm not sure if unifying it will actually improve readability or not.

NautiluX (commented):
/lgtm

bmeng:
@robotmaxtron @rogbas @tnierman Could you help take a look and merge this? It had been reviewed by Manuel, and had been tested in my staging cluster.

Thanks.

robotmaxtron (approved):
/approve

PR #171, opened 2021-12-08
OSD-8075 - Enforce webhooks via generated namespace lists

tnierman:
(Hopefully) completes: https://issues.redhat.com/browse/OSD-8075

---

Follow up to #168 and #166 

Adds a `Makefile` rule to generate a `PrivilegedNamespaces` list from the relevant configMaps in managed-cluster-config.

Webhooks will be enforced using this list, rather than using regular expressions covering all `openshift-.*|kube-.*|redhat-.*` namespaces.


jharrington22:
/lgtm

@csheremeta @robotmaxtron @rogbas I'd like to see this merged today if you could approve it :)

pkg/config/generate/namespaces.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+var (
+	// Base lists - default values which will always be enforced regardless of managed-cluster-config
+	namespaces = []string{"^redhat.*"}
+	configmaps = []string{}
+)
+
+const (
+	// generatedFileName defines the path to the generated file relative to the invoking go:generate command
```
lisa, line 32: Great comment 👍

jharrington22 (approved):
/lgtm

lisa (approved):
/lgtm

One thing to look into maybe is some Github syntactic sugar to consider the generated files as auto-generated, so they don't show up in the diff list. It's really outside the scope of the change (since these are obviously valid Go files and no problem with the filenames), but 🤷‍♀️ it'll help future reviews.

jharrington22:
Manually approving this PR, its well tested and reviewed.

PR #176, opened 2022-01-18
add validation webhook for scc

bmeng:
add the webhook for scc with configurable allowed user and group

allow the service account for node-exporter by default

tested in live cluster that the webhook works well, and the upgrade can be finished without issue.

fixes: [OSD-9642](https://issues.redhat.com/browse/OSD-9642)

cc @NautiluX @Tessg22 @robotmaxtron

### DIFF
--- a/pkg/webhooks/scc/scc.go
+++ b/pkg/webhooks/scc/scc.go
@@ -36,7 +36,6 @@ var (
 		},
 	}
 	allowedUsers = []string{
-		"system:serviceaccount:openshift-monitoring:node-exporter",
 		"system:serviceaccount:openshift-monitoring:cluster-monitoring-operator",
 	}
 	allowedGroups = []string{}
--- a/pkg/webhooks/scc/scc.go
+++ b/pkg/webhooks/scc/scc.go
@@ -1,0 +1,207 @@
+package scc
+
+import (
+	"fmt"
+	"net/http"
+
+	securityv1 "github.com/openshift/api/security/v1"
+	"github.com/openshift/managed-cluster-validating-webhooks/pkg/webhooks/utils"
+	admissionv1 "k8s.io/api/admission/v1"
+	admissionregv1 "k8s.io/api/admissionregistration/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	admissionctl "sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+const (
+	WebhookName string = "scc-validation"
+	docString   string = `Managed OpenShift Customers may not modify the following default SCCs: %s`
+)
+
+var (
+	timeout int32 = 2
+	log           = logf.Log.WithName(WebhookName)
+	scope         = admissionregv1.ClusterScope
+	rules         = []admissionregv1.RuleWithOperations{
+		{
+			Operations: []admissionregv1.OperationType{"UPDATE", "DELETE"},
+			Rule: admissionregv1.Rule{
+				APIGroups:   []string{"security.openshift.io"},
+				APIVersions: []string{"*"},
+				Resources:   []string{"securitycontextconstraints"},
+				Scope:       &scope,
+			},
+		},
+	}
+	allowedUsers = []string{
+		"system:serviceaccount:openshift-monitoring:node-exporter",
+		"system:serviceaccount:openshift-monitoring:cluster-monitoring-operator",
+	}
+	allowedGroups = []string{}
+	defaultSCCs   = []string{
+		"anyuid",
+		"hostaccess",
+		"hostmount-anyuid",
+		"hostnetwork",
+		"node-exporter",
+		"nonroot",
+		"privileged",
+		"restricted",
+		"pipelines-scc",
+	}
+)
+
+type SCCWebHook struct {
+	s runtime.Scheme
+}
+
+// NewWebhook creates the new webhook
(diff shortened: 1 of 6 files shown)


pkg/webhooks/scc/scc.go
```diff
@@ -0,0 +1,212 @@
...
+		return nil, err
+	}
+
+	return scc, nil
+}
+
+// isAllowedUserGroup checks if the user or group is allowed to perform the action
+func isAllowedUserGroup(request admissionctl.Request) bool {
+	isAllowedUser := false
+	if len(allowedUsers) > 0 {
```
NautiluX, line 125: why do we need this check?

pkg/webhooks/scc/scc.go
```diff
@@ -0,0 +1,212 @@
...
+		if utils.SliceContains(request.UserInfo.Username, allowedUsers) {
+			isAllowedUser = true
+		}
+	}
+
+	isAllowedGroup := false
+	if len(allowedGroups) > 0 {
```
NautiluX, line 132: why do we need this check?

pkg/webhooks/scc/scc.go
```diff
@@ -0,0 +1,212 @@
...
+
+	return scc, nil
+}
+
+// isAllowedUserGroup checks if the user or group is allowed to perform the action
+func isAllowedUserGroup(request admissionctl.Request) bool {
+	isAllowedUser := false
+	if len(allowedUsers) > 0 {
+		if utils.SliceContains(request.UserInfo.Username, allowedUsers) {
+			isAllowedUser = true
```
NautiluX, line 127: why not return true immediately?

pkg/webhooks/scc/scc.go
```diff
@@ -0,0 +1,212 @@
...
+	isAllowedGroup := false
+	if len(allowedGroups) > 0 {
+		for _, group := range allowedGroups {
+			if utils.SliceContains(group, request.UserInfo.Groups) {
+				isAllowedGroup = true
+			}
+		}
+	}
+
+	return (isAllowedUser || isAllowedGroup)
```
NautiluX, line 140: I think this function could be simplified to something like
```suggestion
	if utils.SliceContains(request.UserInfo.Username, allowedUsers) {
		return true
	}

	for _, group := range allowedGroups {
		if utils.SliceContains(group, request.UserInfo.Groups) {
			return true
		}
	}
	return false
```

pkg/webhooks/scc/scc.go
```diff
@@ -0,0 +1,212 @@
...
+			return ret
+		}
+	case admissionv1.Update:
+		if isDefaultSCC(scc) && !isAllowedUserGroup(request) {
+			log.Info(fmt.Sprintf("Updating operation detected on default SCC: %v", scc.Name))
+			ret = admissionctl.Denied(fmt.Sprintf("Modifying default SCCs %v is not allowed", defaultSCCs))
+			ret.UID = request.AdmissionRequest.UID
+			return ret
+		}
+	}
```
NautiluX, line 97: Code duplication is growing now... can we at least move the if outside the switch?
```suggestion
	if isDefaultSCC(scc) && !isAllowedUserGroup(request) {
		switch request.Operation {
		case admissionv1.Delete:
			log.Info(fmt.Sprintf("Deleting operation detected on default SCC: %v", scc.Name))
			ret = admissionctl.Denied(fmt.Sprintf("Deleting default SCCs %v is not allowed", defaultSCCs))
			ret.UID = request.AdmissionRequest.UID
			return ret
		case admissionv1.Update:
			log.Info(fmt.Sprintf("Updating operation detected on default SCC: %v", scc.Name))
			ret = admissionctl.Denied(fmt.Sprintf("Modifying default SCCs %v is not allowed", defaultSCCs))
			ret.UID = request.AdmissionRequest.UID
			return ret
		}
	}
```

NautiluX (changes requested):
Looks good overall, but I think a bit of refactoring could simplify it and remove a bit of code duplication.

bmeng:
thanks @NautiluX I have fixed the comments you made, and tested pass on a live cluster.

Tessg22:
As explained by Christoph in https://issues.redhat.com/browse/OSD-9642, it is not the node_exporter service account that should be allowed to bypass the webhook. I am going to double check that it is CVO that is updating node_exporter SCC and update the card description accordingly.

pkg/webhooks/scc/scc.go
```diff
@@ -0,0 +1,207 @@
...
+	defaultSCCs   = []string{
+		"anyuid",
+		"hostaccess",
+		"hostmount-anyuid",
+		"hostnetwork",
+		"node-exporter",
+		"nonroot",
+		"privileged",
+		"restricted",
+		"pipelines-scc",
```
tnierman, line 52: Should the `splunkforwarder` SCC be added to this list?

bmeng: the default here covers the default SCCs created by OCP I think. the splunkforwarder was added by security team, and also it is not mentioned in the card OSD-8377.
It should not be a blocker here. And we can add it later if needed.

pkg/webhooks/scc/scc.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+
+	for _, group := range allowedGroups {
+		if utils.SliceContains(group, request.UserInfo.Groups) {
+			return true
+		}
+	}
+
+	return false
+}
```
NautiluX, line 136: that looks nice now :)

NautiluX (commented):
the code looks quite clean now, thanks for updating. Once we validate the right SA is in the allowlist, I'm happy with the PR :)

NautiluX:
As documented by the ticket @bmeng verified upgrades are working with this user.
/unhold

PR #182, opened 2022-03-14
Adding ocmagent.managed.openshift.io API group for regular user webhook

ravitri:
As per [OSD-10104](https://issues.redhat.com/browse/OSD-10104), adding API group `ocmagent.managed.openshift.io` to deny non SRE users to modify resources like `OcmAgent` and `ManagedNotifications`.

Remarks for the reviewer:
1. Following commands were run after adding the API group in `regularuser.go`:
  - make syncset
  - make generate
  - make docs > docs/webhooks.json && make DOCFLAGS=-hideRules docs > docs/webhooks-short.json
2. Not quite sure why but the `make generate` command added the following entry in `pkg/config/namespaces.go` file too. For now, have added that change as part of this PR:
```bash
"^redhat-openshift-connectors$"
```
3. The change was tested against the cluster and a cluster-admin user (non SRE user) couldn't change the resources in the added group.


NautiluX (commented):
/lgtm

NautiluX:
AFAIK `make generate` will pull a new namespace list from [managed-cluster-config](https://github.com/openshift/managed-cluster-config/blob/master/resources/managed/all-osd-resources.yaml) but when something is added there, it should be updated here additionally using a separate PR.

ravitri:
@NautiluX - Thanks a lot! Have removed the change for the `make generate` command and will include it in separate PR.

PR #184, opened 2022-03-16
Adds a better error message to node updates

iamkirkbater:
This PR aims to update the error message when a cluster admin attempts
to edit a node in order to add labels/taints, guiding them to use OCM or
the ROSA CLI instead on the MachinePool.

This satisfies OSD-9887.

To Test:
Build and push image to cluster.
Attempt to label or taint a node as cluster-admin.
Validate you get the new error message.


pkg/webhooks/regularuser/regularuser.go
```diff
@@ -221,12 +223,26 @@ func (s *RegularuserWebhook) authorized(request admissionctl.Request) admissionc
...
 	log.Info("Denying access", "request", request.AdmissionRequest)
 	ret = admissionctl.Denied("Prevented from accessing Red Hat managed resources. This is in an effort to prevent harmful actions that may cause unintended consequences or affect the stability of the cluster. If you have any questions about this, please reach out to Red Hat support at https://access.redhat.com/support")
 	ret.UID = request.AdmissionRequest.UID
 	return ret
 }
 
+// isProtectedActionOnNode checks if the request is to update a node
+func isProtectedActionOnNode(request admissionctl.Request) bool {
+	return request.Kind.Kind == nodeKind &&
+		string(request.Operation) == nodeAction
```
lisa, line 243: I think it's okay to go without the constants since they're only used here.

lisa (changes requested):
Please provide unit test(s) around the new behaviour that shows how a variety of users are impacted.

pkg/webhooks/regularuser/regularuser.go
```diff
@@ -221,12 +221,26 @@ func (s *RegularuserWebhook) authorized(request admissionctl.Request) admissionc
...
 	log.Info("Denying access", "request", request.AdmissionRequest)
 	ret = admissionctl.Denied("Prevented from accessing Red Hat managed resources. This is in an effort to prevent harmful actions that may cause unintended consequences or affect the stability of the cluster. If you have any questions about this, please reach out to Red Hat support at https://access.redhat.com/support")
 	ret.UID = request.AdmissionRequest.UID
 	return ret
 }
 
+// isProtectedActionOnNode checks if the request is to update a node
+func isProtectedActionOnNode(request admissionctl.Request) bool {
+	return request.Kind.Kind == "Node" &&
+		string(request.Operation) == "UPDATE"
```
lisa, line 241:
```suggestion
	return request.Kind.Kind == "Node" && request.Operation == admissionv1.Update
```

We can compare to the upstream type and avoid coercing to a string.

lisa (changes requested):
One minor thing about type comparison. Looks good otherwise.

lisa (approved):

PR #189, opened 2022-05-23
OSD-10922 - Fix webhook to block TechPreviewNoUpgrade

Tafhim:
The [OSD-10922](https://issues.redhat.com//browse/OSD-10922) ticket was resolved. But there were some bugs observed on follow up testing.

- The API Groups needed to be "config.openshift.io" since that is the API Group for FeatureGates
- The *Delete* operation does not really apply here since *TechPreviewNoUpgrade* is a featureSet not a *Feature Gate*. It is only implemented as a FeatureSet on a new or an existing Feature Gate. Thus, only testing Create and Update operations should suffice. A note on [Openshift Feature Gate page](https://docs.openshift.com/container-platform/4.8/nodes/clusters/nodes-cluster-enabling-features.html) clearly says that deletion of TechPreviewNoUpgrade is not supported anyway. And also, as I have tested manually both in Upgrade and Delete scenarios, the FeatureSet cannot be taken off once it's implemented.
- The Scope of the webhook should be *Cluster*, not *Namespaced*.

Those issues have been resolved in this ticket.


pkg/webhooks/techpreviewnoupgrade/techpreviewnoupgrade.go
```diff
@@ -103,23 +101,31 @@ func (s *TechPreviewNoUpgradeWebhook) renderFeatureGate(request admissionctl.Req
...
 func (s *TechPreviewNoUpgradeWebhook) authorized(request admissionctl.Request) admissionctl.Response {
+	var ret admissionctl.Response
+
 	featureGate, err := s.renderFeatureGate(request)
 
 	if err != nil {
-		return admissionctl.Errored(http.StatusBadRequest, err)
+		log.Error(err, "Couldn't render a FeatureGate from the incoming request")
+
+		ret = admissionctl.Errored(http.StatusOK, err)
```
NautiluX, line 111: Why does this need to change from StatusBadRequest to StatusOK?

Tafhim: Because the request is not really malformed or bad in itself. The operation is not authorized to be performed. 
I am more open to ideas in this, should we instead use `StatusForbidden`? @NautiluX

NautiluX: I'm still not sure what you mean by the operation is not authorized, doesn't that error mean the request couldn't be parsed into a FeatureGate? So wouldn't that be a bad payload rather than a forbidden request?

Tafhim: Oh I was thinking in the sense that this response is being generated from the `authorised` method. When this block of code executes, the request is not authorized since the error response is returned. 

But I agree with you now that I realize that this is meant for requests not having a well formed FeatureGate. I will restore it back to `StatusBadRequest`.

Tafhim: @NautiluX I have reverted this to `StatusBadRequest` as per discussion above.

Tafhim:
Hi @robotmaxtron can I get an approved on this? It already has lgtm but I needs approval. :)

PR #191, opened 2022-06-06
Update documentation files with document string fix for TechPreviewNo…

Tafhim:
The TechPreviewNoUpgrade webhook document string was misleading. This PR corrects the issue.


Tafhim:
@robotmaxtron Hi Max, it's a really small change in the documentation string. Would you mind approving it? Already has an lgtm.

PR #192, opened 2022-06-23
Block modification of machineconfig/machineconfigpools

cblecker:


pkg/webhooks/regularuser/regularuser.go
```diff
@@ -76,6 +76,15 @@ var (
...
 		},
+		{
+			Operations: []admissionregv1.OperationType{"*"},
+			Rule: admissionregv1.Rule{
+				APIGroups:   []string{"machineconfiguration.openshift.io"},
+				APIVersions: []string{"*"},
+				Resources:   []string{"machineconfigs", "machineconfigpools"},
+				Scope:       &scope,
+			},
+		},
```
lisa, line 87: @cblecker This seems similar in scope to what was being done on line 45. For context:

```go
{
			Operations: []admissionregv1.OperationType{"*"},
			Rule: admissionregv1.Rule{
				APIGroups: []string{
					"cloudcredential.openshift.io",
					"machine.openshift.io",
					"admissionregistration.k8s.io",
					// Deny ability to manage SRE resources
					// oc get --raw /apis | jq -r '.groups[] | select(.name | contains("managed")) | .name'
					"addons.managed.openshift.io",
					"cloudingress.managed.openshift.io",
					"managed.openshift.io",
					"ocmagent.managed.openshift.io",
					"splunkforwarder.managed.openshift.io",
					"upgrade.managed.openshift.io",
				},
				APIVersions: []string{"*"},
				Resources:   []string{"*/*"},
				Scope:       &scope,
			},
		},
```

For the sake of consolidating all of the "machine-related business" in one place, does it make sense to move line 45 to this new block?

lisa: Whoops! My mistake. I missed the specified resources in this (new) block.

cblecker: Thanks for the review! Wanted to make sure I wasn't missing something :)

robotmaxtron (approved):

PR #196, opened 2022-09-14
OSD-8377 - Add webhook to block proxy and user-ca-bundle config map modification

Tafhim:
OCM has introduced functionality to configure Proxy and Additional Trust Bundle properly. We now want to block anyone manually configuring these resources.


pkg/webhooks/regularuser/regularuser.go
```diff
@@ -307,6 +322,25 @@ func isNetNamespaceValid(s *RegularuserWebhook, request admissionctl.Request) bo
...
+	if err != nil {
+		return false
+	}
+	configMap := &corev1.ConfigMap{}
+	err = decoder.DecodeRaw(request.Object, configMap)
+	if err != nil {
+		return false
+	}
+
+	if configMap.ObjectMeta.Name != "user-ca-bundle" || configMap.ObjectMeta.Namespace != "openshift-config" {
```
feichashao, line 337: Is it flexible to make "user-ca-bundle" and "openshift-config" as a var list like `Resources:   []string{"clusterversions", "clusterversions/status", "schedulers", "apiservers", "proxies"},` above?

feichashao: other functions also doing the similar with hard-coded strings. I feel ok about this.

robotmaxtron (approved):
/approve

PR #203, opened 2023-01-27
 Remove autoscaler from admissionwebhook ruleset 

gburges:
This makes ROSA/OSD cluster admins be able to configure clusterAutoScaler

### DIFF
--- a/build/selectorsyncset.yaml
+++ b/build/selectorsyncset.yaml
@@ -327,16 +327,6 @@ objects:
           - '*/*'
           scope: '*'
         - apiGroups:
-          - autoscaling.openshift.io
-          apiVersions:
-          - '*'
-          operations:
-          - '*'
-          resources:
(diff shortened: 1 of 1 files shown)


fahlmant (changes requested):
OSD/ROSA customers are intended to control the autoscaler through OCM, not directly on cluster. Please open an OSD card as an RFE for the specific use case you're looking for

gburges:
https://issues.redhat.com/browse/OHSS-18212
https://issues.redhat.com/browse/SDE-1558

gburges:
As far as i know, in OCM clusterAutoscaler can only be enabled/disabled, not configured. I'd like to test out different options, change scale TH, ignore daemonsets, turn off balancing, and so on.

fahlmant:
@gburges That's intentional, and we'll need an RFE (that you linked) accepted before we can change this webhook

PR #206, opened 2023-02-15
Minor improvements to Hypershift docs

mrbarge:
This is a small PR that adds more detail to the way the Hypershift deployment of MCVW works, specifically:
- explicitly stating that the validating webhooks are part of the PKO package
- explaining how the continuous rollout happens.


mjlshen (commented):
/lgtm

PR #212, opened 2023-04-17
[Osd 15333 ] PKO to Inject caBundle of Management cluster to the Hosted cluster

a7vicky:
This PR is using new api of PKO to parameterizing Injection of caBundle of Management cluster to the Hosted cluster resource `ValidatingWebhookConfiguration`
Ref: https://package-operator.run/docs/concepts/object-template/

- Copy configmap `openshift-service-ca.crt`  value of `.data.service-ca\.crt` and inject into hosted cluster `ValidatingWebhookConfiguration`   `clientConfig.caBundle` which is deployed through PKO

### DIFF
--- a/config/package/manifest.yaml
+++ b/config/package/manifest.yaml
@@ -6,11 +6,20 @@ spec:
   scopes:
     - Namespaced
   phases:
+    - name: config
+    - name: rbac
+    - name: deploy
     - name: webhooks
       class: hosted-cluster
-    - name: rbac
-    - name: config
-    - name: deploy
+  config:
+    openAPIV3Schema:
+      properties:
+        serviceca:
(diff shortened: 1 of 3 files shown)


mrbarge:
@a7vicky we'll also need to add this to the `config/package/manifest.yaml`

```
  config:
    openAPIV3Schema:
      properties:
        serviceca:
          description: Service Certificate Authority to do ....
          type: string
      required:
      - serviceca
      type: object
```

mrbarge:
/lgtm

I have tested this alongside @a7vicky and am confident that it is all good.

The hold has been placed so that we can unhold it at a time of our preference and watch it hit HyperShift staging clusters live.

PR #213, opened 2023-04-23
Create new validating webhook for icp/icsp 

mjlshen:
[ImageContentPolicy](https://docs.openshift.com/container-platform/4.12/rest_api/config_apis/imagecontentpolicy-config-openshift-io-v1.html) and [ImageContentSourcePolicy](https://docs.openshift.com/container-platform/4.12/rest_api/operator_apis/imagecontentsourcepolicy-operator-openshift-io-v1alpha1.html) resources allow
customers to configure repository mirrors. This validating webhook will
prevent configuring mirrors for quay.io, registry.redhat.io, and
registry.access.redhat.com repositories as those will likely have an
adverse affect on managed OpenShift.

### DIFF
--- a/pkg/webhooks/imagecontentpolicies/imagecontentpolicies.go
+++ b/pkg/webhooks/imagecontentpolicies/imagecontentpolicies.go
@@ -1,0 +1,194 @@
+package imagecontentpolicies
+
+import (
+	"fmt"
+	"net/http"
+	"regexp"
+
+	"github.com/go-logr/logr"
+	configv1 "github.com/openshift/api/config/v1"
+	operatorv1alpha1 "github.com/openshift/api/operator/v1alpha1"
+	"github.com/openshift/managed-cluster-validating-webhooks/pkg/webhooks/utils"
+	admissionv1 "k8s.io/api/admission/v1"
+	admissionregv1 "k8s.io/api/admissionregistration/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+const (
+	WebhookName = "imagecontentpolicies-validation"
+	WebhookDoc  = "Managed OpenShift customers may not create ImageContentPolicy or ImageContentSourcePolicy resources that configure mirrors for quay.io, registry.redhat.com, nor registry.access.redhat.com."
+	// unauthorizedRepositoryMirrors is a regex that is used to reject certain specified repository mirrors.
+	// Generally all contained regexes follow a similar pattern, i.e. rejecting quay.io or quay.io/.*
+	unauthorizedRepositoryMirrors = `(^quay\.io(/.*)?$|^registry\.redhat\.io(/.*)?$|^registry\.access\.redhat\.com(/.*)?)`
+)
+
+type ImageContentPoliciesWebhook struct {
+	scheme *runtime.Scheme
+	log    logr.Logger
+}
+
+func NewWebhook() *ImageContentPoliciesWebhook {
+	return &ImageContentPoliciesWebhook{
+		scheme: runtime.NewScheme(),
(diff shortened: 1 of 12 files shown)


pkg/webhooks/imagecontentpolicies/imagecontentpolicies.go
```diff
@@ -0,0 +1,188 @@
...
+	return false
+}
+
+func (w *ImageContentPoliciesWebhook) Name() string {
+	return WebhookName
+}
+
+func (w *ImageContentPoliciesWebhook) FailurePolicy() admissionregv1.FailurePolicyType {
+	// Fail-closed
+	return admissionregv1.Fail
```
mjlshen, line 102: Most other webhooks I've seen in this repo are `admissionregv1.Ignore` i.e. "fail-open".

There are definitely cons to fail-closed if this webhook is failing, however is it better if we fail-open and a problematic`ImageContentPolicy` gets created is that preferred? I was leaning towards fail-closed to start out with and we can move it to fail-open if UX concerns arise and as we add more observability into how these CRs interact with existing alerts?

pkg/webhooks/imagecontentpolicies/imagecontentpolicies.go
```diff
@@ -0,0 +1,188 @@
...
+	icp := configv1.ImageContentPolicy{}
+	icsp := operatorv1alpha1.ImageContentSourcePolicy{}
+	allowedResp := admission.Response{
+		AdmissionResponse: admissionv1.AdmissionResponse{
+			UID:     request.UID,
+			Allowed: true,
+		},
+	}
+
+	if len(request.Object.Raw) > 0 {
```
lisa, line 54: What is the importance of when the length of the raw object is greater than 0? It looks like we're Doing Things ™️ when it is the case, but when it is not greater than 0, the request is allowed by virtue of lines [47-52](https://github.com/openshift/managed-cluster-validating-webhooks/pull/213/files#diff-9697b102dd57fb098f8ff05ef470c690297d5f5fe218980f456b7b4e7cc0a02dR48-R52)?

pkg/webhooks/imagecontentpolicies/imagecontentpolicies.go
```diff
@@ -0,0 +1,188 @@
...
+
+	return false
+}
+
+func (w *ImageContentPoliciesWebhook) Name() string {
+	return WebhookName
+}
+
+func (w *ImageContentPoliciesWebhook) FailurePolicy() admissionregv1.FailurePolicyType {
+	// Fail-closed
```
lisa, line 101: Why are we failing closed? This is different than other hooks, as @mjlshen points out below. Please expand on why we need to fail closed so others coming to modify this hook have the context at hand.

pkg/webhooks/imagecontentpolicies/imagecontentpolicies.go
mjlshen: The length should never be 0 for a "valid" request since it means there's no object in the request. It's more of a guard before we attempt to check `request.RequestKind.Kind` or importantly before we attempt to `decoder.Decode(request, &kind)`

pkg/webhooks/imagecontentpolicies/imagecontentpolicies.go
mjlshen: Moved my PR comment into a code comment :D

pkg/webhooks/imagecontentpolicies/imagecontentpolicies.go
lisa: Should this logic check move to `Validate`?

pkg/webhooks/imagecontentpolicies/imagecontentpolicies.go
mjlshen:
```
	// Fail-closed because if we allow a problematic ImageContentPolicy/ImageContentSourcePolicy through,
	// it will have significant impact on the cluster. We should not modify this to fail-open unless we have
	// other specific observability and guidance to detect misconfigured ImageContentPolicy/ImageContentSourcePolicy
	// resources.
```

PR #215, opened 2023-04-28
Fix bugs where the icp webhook was returning invalid responses

mjlshen:
This commit fixes bugs in and improves the newly created icp webhook:

1. All allowed and denied responses must contain the request's UID. Previously only allowed responses contained the requests's UID.
2. All allowed responses must return an HTTP 200. Previously the code was unset.
3. Adds an info-level log message when the webhook denies a request.

I created a reusable utils.WebhookResponse() function that can help ease the understanding gap for new contributors to this repo and added test cases that would have caught the bugs in the initial creation of this webhook.

OSD-15757 

Related to #213 where this webhook was first created


NautiluX (commented):
/lgtm

PR #217, opened 2023-05-02
Continue blocking registry.redhat.io exactly, but allow mirroring otherwise

mjlshen:
Due to recent findings in OSD-15757, teams are mirroring specific repos under registry.redhat.io


mjlshen:
/unhold

To unblock teams for now

PR #218, opened 2023-05-03
Unit test icp webhook validation and authorization

mjlshen:
This PR adds unit tests that would have caught the bugs that were fixed in #215 and #216

Of note is that GVK's need not be registered with the scheme because it marshals the object into `unstructured.Unstructured` under the hood and it passes the unit test!


NautiluX (commented):
/lgtm

PR #219, opened 2023-05-16
OSD-15333 Set caBundle to PKO templating in generated resources

mrbarge:
# What it does
As part of OSD-15333 to deprecate the `caBundle` patching cronjob in Hypershift, we need to specify to `package-operator` (PKO) that it needs to set the `caBundle` in webhooks when applying them to hosted clusters.

We'd previously done this in the PKO template (manually) in a prior PR, but forgot that this file is auto-generated by the build process, so the next time a change came along, it was lost.

# Notes
However, it's not quite easy to set the `caBundle` to free-form text because its type is a `[]byte` so when it gets marshalled to json, it turns into base64. To have it as free-form text so that PKO can interpret it, we have to intercept the Webhook marshalling to JSON when building the PKO package and override the `caBundle` value. Which is part of what this change does.

# Refs
OSD-15333


mrbarge:
/hold
whilst it is tested

mrbarge:
Testing successfully
/hold cancel

PR #221, opened 2023-05-29
Split regularuser webhook into universal and OSD variants

mrbarge:
# What it does
Splits the existing `sre-regular-user-validation` webhook into two separate ones:

- `sre-regular-user-validation` (existing) - which contains regular-user validations that are universal across OSD/ROSA/Hypershift
- `sre-regular-user-validation-osd` (new) - which contains regular-user validations that are specific to OSD/ROSA

Currently, the `sre-regular-user-validation-osd` webhook being introduced only contains one validation:
- regular users cannot update nodes
This has been removed from `sre-regular-user-validation` accordingly.

# Why
The `sre-regular-user-validation` webhook as-is contains validations which are not suitable for Hypershift - specifically the guard against updating `node`s, which has been deemed acceptable in Hypershift. 

Presently the workaround for this has been to disable _all_ regular-user protections in Hypershift, which is not ideal.

As there will likely be potentially other things we want to safeguard in OSD/ROSA but not in Hypershift, that is why this split-out has occurred.

Other implementation options were considered, but ultimately not used:
- We could trim what `rules` the existing `sre-regular-user-validation` webhook contained in Hypershift, but then the webhook's documentation does not align with how it actually runs in Hypershift, and additionally doing this in a non-hacky way was going to be very ugly.

# Jira
[OSD-15334](https://issues.redhat.com//browse/OSD-15334)

# Testing
This has been successfully tested on a OSD cluster by:
- Ensuring that `cluster-admins` and `dedicated-admins` still cannot update nodes with the new webhook in place.
- Ensuring that if the new webhook is deleted, that `cluster-admins` can update nodes.
Remaining testing validation will be obtained via osde2e.


lisa:
@mrbarge I'd love to see some comments in the sub-packages that call out the purpose of the `osd` variant. Also, there's a reference in [README.md](/README.md) to the old regularuser package that should be updated (this is relatively minor, all things considered).

In spirit, this is a well reasoned change. I'll let others discuss the individual changes in code.

mrbarge:
@lisa Thank you for the review, all your suggestions are very helpful! I've aimed to address them in the latest commit.

NautiluX (commented):
/lgtm

PR #222, opened 2023-06-03
OSD-16717 remove validation-webhook service account

mrbarge:
# What is this PR?
Removes the:
- `validation-webhook` ServiceAccount
- `webhook-validation-cr` ClusterRole
- `webhook-validation` ClusterRoleBinding
- the use of the `validation-webhook` in the OSD/ROSA DaemonSet and HCP Deployment

# Why?
Before SRE had backplane we had to check group membership within a webhook.  The RBAC for this was bound at a ClusterRoleBinding.  It is still in place today.  For HCP the SA bound to the webhook pods still exists and projects the SA token but in this context (HCP) there is no RBAC assigned.

There is no reason for SRE webhooks to have KAS access anymore and it should be removed.

# Refs
OSD-16717

### DIFF
--- a/config/package/resources.yaml.gotmpl
+++ b/config/package/resources.yaml.gotmpl
@@ -1,13 +1,5 @@
 ---
 apiVersion: v1
-kind: ServiceAccount
-metadata:
-  annotations:
-    package-operator.run/phase: rbac
-  creationTimestamp: null
-  name: validation-webhook
----
-apiVersion: v1
 kind: ConfigMap
 metadata:
   annotations:
@@ -117,7 +109,6 @@ spec:
           name: service-ca
           readOnly: true
       restartPolicy: Always
-      serviceAccountName: validation-webhook
       tolerations:
       - effect: NoSchedule
         key: hypershift.openshift.io/hosted-control-plane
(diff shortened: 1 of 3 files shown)


config/package/resources.yaml.gotmpl
```diff
@@ -162,7 +153,8 @@ webhooks:
     - CREATE
     - UPDATE
     resources:
-    - imagecontentpolicies
+    - imagedigestmirrorsets
```
mrbarge, line 156: Just noting that this seems to be a consequence of PR #221 merging after PR #220 and having the old HS package contents in it. Which seems a problem that needs to be detected in future by `make` validation. I will create a follow-up ticket to track that.

But to summarize - this change is expected and fine.

cblecker:
Will this/should we clean up the existing SA/rbac from existing clusters?

mrbarge:
> Will this/should we clean up the existing SA/rbac from existing clusters?

My thought on this has been..  hopefully yes, because the SelectorSyncSet has a mode of `Sync`, so they _should_ go. If they don't, I'll create a follow-up MCC pruning task to get rid of them before closing this out.

cblecker (approved):

PR #224, opened 2023-06-07
OSD-16762 allow system:admin to update SCCs

mrbarge:
# What it does

Allow the `system:admin` user to be able to UPDATE the default set of OpenShift SCCs.

# Why?

The `cluster-version-operator` in HCP namespaces uses `system:admin` to patch SCCs during upgrades, rather than the regular OCP user of `system:serviceaccount:openshift-monitoring:cluster-monitoring-operator`.  This then gets blocked by MCVW and is causing y-stream upgrades to fail.

# Refs

[OSD-16762](https://issues.redhat.com//browse/OSD-16762)


NautiluX (commented):
/lgtm

PR #225, opened 2023-06-09
Don't project SA token

jewzaam:
https://issues.redhat.com/browse/OSD-16717

### DIFF
--- a/config/package/resources.yaml.gotmpl
+++ b/config/package/resources.yaml.gotmpl
@@ -54,6 +54,7 @@ spec:
       labels:
         app: validation-webhook
     spec:
+      automountServiceAccountToken: "false"
       affinity:
         nodeAffinity:
           preferredDuringSchedulingIgnoredDuringExecution:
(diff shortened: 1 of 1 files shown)


config/package/resources.yaml.gotmpl
```diff
@@ -95,7 +96,7 @@ spec:
         - -cacert
         - /service-ca/service-ca.crt
         - -tls
-        image: REPLACED_BY_PIPELINE
+        image: quay.io/app-sre/managed-cluster-validating-webhooks:4c7a8ed
```
mrbarge, line 99: We shouldn't need to change this line, it gets changed during the app interface CI build to reflect the latest commit.

config/package/resources.yaml.gotmpl
```diff
@@ -54,6 +54,7 @@ spec:
       labels:
         app: validation-webhook
     spec:
+      automountServiceAccountToken: "false"
```
mrbarge, line 57: It complains that this is not a boolean due to the quotes.

```suggestion
      automountServiceAccountToken: false
```

mrbarge:
I'm doing this in #226 instead, as the change has been made in the wrong place.
Recommend we close this.

PR #227, opened 2023-06-14
Update ICP webhook documentation message

mjlshen:
It was unclear that customers are able to configure mirrors for specific repositories and only blocked from configuring mirrors across the entirety of certain domains.


NautiluX (commented):
/lgtm

PR #228, opened 2023-06-15
OSD-16469 Block ingress controller provisioning on master and infra

Tafhim:
This PR adds a webhook that inspects ingress controller creation request from customers. It blocks these requests when they have tolerations for infra and master nodes.

Issue: [OSD-16469](https://issues.redhat.com//browse/OSD-16469)

### DIFF
--- a/pkg/webhooks/ingresscontroller/ingresscontroller.go
+++ b/pkg/webhooks/ingresscontroller/ingresscontroller.go
@@ -1,0 +1,169 @@
+package ingresscontroller
+
+import (
+	"fmt"
+	"net/http"
+	"regexp"
+	"strings"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+	"github.com/openshift/managed-cluster-validating-webhooks/pkg/webhooks/utils"
+	admissionregv1 "k8s.io/api/admissionregistration/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	admissionctl "sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+const (
+	WebhookName   string = "ingresscontroller-validation"
+	docString     string = `Managed OpenShift Customer may create IngressControllers without necessary taints. This can cause those workloads to be provisioned on infra or master nodes.`
+	allowedGroups string = `^system:serviceaccounts:(kube.*|openshift.*|default|redhat.*|osde2e-[a-z0-9]{5})`
+)
+
+var (
+	log   = logf.Log.WithName(WebhookName)
+	scope = admissionregv1.NamespacedScope
+	rules = []admissionregv1.RuleWithOperations{
+		{
+			Operations: []admissionregv1.OperationType{admissionregv1.Create, admissionregv1.Update},
+			Rule: admissionregv1.Rule{
+				APIGroups:   []string{"operator.openshift.io"},
+				APIVersions: []string{"*"},
+				Resources:   []string{"ingresscontroller", "ingresscontrollers"},
+				Scope:       &scope,
+			},
+		},
+	}
+	allowedUsers = []string{
+		"backplane-cluster-admin",
+	}
+	allowedGroupsRe = regexp.MustCompile(allowedGroups)
+)
+
+type IngressControllerWebhook struct {
+	s runtime.Scheme
+}
+
+// ObjectSelector implements Webhook interface
+func (wh *IngressControllerWebhook) ObjectSelector() *metav1.LabelSelector { return nil }
+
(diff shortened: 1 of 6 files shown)


pkg/webhooks/ingresscontroller/ingresscontroller.go
```diff
@@ -0,0 +1,162 @@
...
+}
+
+func (wh *IngressControllerWebhook) authorized(request admissionctl.Request) admissionctl.Response {
+	var ret admissionctl.Response
+	ic, err := wh.renderIngressController(request)
+	if err != nil {
+		log.Error(err, "Couldn't render an IngressController from the incoming request")
+		return admissionctl.Errored(http.StatusBadRequest, err)
+	}
+	if ic.Namespace != "openshift-ingress-operator" {
```
bmeng, line 115: I am not sure if this required, as the operator itself will reconcile the `ingresscontroller` within the namespace only, so created outside should be harmless but doing nothing.

pkg/webhooks/ingresscontroller/ingresscontroller.go
```diff
@@ -0,0 +1,162 @@
...
+		ret = admissionctl.Denied("Not allowed to create or update an ingress controller in namespace other than openshift-ingress-operator")
+		ret.UID = request.AdmissionRequest.UID
+
+		return ret
+	}
+
+	if ic.Name == "default" {
```
bmeng, line 122: This as well, the default ingresscontroller should be managed by tools. I am not quite sure if this is needed.

pkg/webhooks/ingresscontroller/ingresscontroller.go
```diff
@@ -0,0 +1,162 @@
...
+)
+
+const (
+	WebhookName string = "ingresscontroller-validation"
+	docString   string = `Managed OpenShift Customer may create IngressControllers without necessary taints. This can cause those workloads to be provisioned on infra or master nodes.`
+)
+
+var (
+	log   = logf.Log.WithName(WebhookName)
+	scope = admissionregv1.ClusterScope
```
bmeng, line 26: This should be namespace scope?

pkg/webhooks/ingresscontroller/ingresscontroller.go
```diff
@@ -0,0 +1,162 @@
...
+	rules = []admissionregv1.RuleWithOperations{
+		{
+			Operations: []admissionregv1.OperationType{admissionregv1.OperationAll},
```
bmeng, line 29: We should limited to CREATE/UPDATE/PATCH here.

pkg/webhooks/ingresscontroller/ingresscontroller.go
bmeng: Also we will need to allow the default ingresscontroller to be created by special user/group, so that it can be created after sync from ocm/hive.

pkg/webhooks/ingresscontroller/ingresscontroller.go
```diff
@@ -0,0 +1,161 @@
...
+func (wh *IngressControllerWebhook) authorized(request admissionctl.Request) admissionctl.Response {
+	var ret admissionctl.Response
+	ic, err := wh.renderIngressController(request)
+	if err != nil {
+		log.Error(err, "Couldn't render an IngressController from the incoming request")
+		return admissionctl.Errored(http.StatusBadRequest, err)
+	}
+
+	// Check if the group does not have exceptions
+	if !isAllowedUserGroup(request) {
```
hectorakemp, line 115: Maybe also a loop here over `nodeSelector` labels?
I.e. reject yaml like this:
```
      nodeSelector:
        node-role.kubernetes.io/infra: "" #or master
      tolerations:
        - effect: NoSchedule
          key: node-role.kubernetes.io/infra
          operator: Exists
```

hectorakemp (changes requested):
99% there - amazing work @Tafhim in such a short space of time. One small addition to the authorize loop, and then I think we're there.

pkg/webhooks/ingresscontroller/ingresscontroller.go
hectorakemp: Resolving from discussion here: https://redhat-internal.slack.com/archives/C0326L38PEH/p1686795999616689

bmeng, line 115: will this block the operators to update the ingresscontroller?

bmeng:
/hold 

pkg/webhooks/ingresscontroller/ingresscontroller.go
hectorakemp: I don't think it will @bmeng, because operators don't ever deploy workloads to non-infra or master nodes, do they?

hectorakemp: I don't think we do @Tafhim - if all three of those operators are able to create ICs, then we're good

hectorakemp (approved):

PR #229, opened 2023-06-15
[OSD-16224] add webhook to prevent critical serviceaccount deletion

bmeng:
Add the webhook for critical serviceccount deletion prevention
update the testutils to include the namespace value in the request

### DIFF
--- a/pkg/webhooks/serviceaccount/serviceaccount.go
+++ b/pkg/webhooks/serviceaccount/serviceaccount.go
@@ -1,0 +1,221 @@
+package serviceaccount
+
+import (
+	"fmt"
+	"net/http"
+	"os"
+
+	admissionv1 "k8s.io/api/admission/v1"
+	admissionregv1 "k8s.io/api/admissionregistration/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	admissionctl "sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
+	"github.com/openshift/managed-cluster-validating-webhooks/pkg/config"
+	"github.com/openshift/managed-cluster-validating-webhooks/pkg/webhooks/utils"
+)
+
+const (
+	WebhookName string = "serviceaccount-validation"
+	docString   string = `Managed OpenShift Customers may not delete the service accounts under the managed namespaces。`
+)
+
+var (
+	timeout int32 = 2
+	log           = logf.Log.WithName(WebhookName)
+	scope         = admissionregv1.NamespacedScope
+	rules         = []admissionregv1.RuleWithOperations{
+		{
+			Operations: []admissionregv1.OperationType{"DELETE"},
+			Rule: admissionregv1.Rule{
+				APIGroups:   []string{""},
+				APIVersions: []string{"v1"},
+				Resources:   []string{"serviceaccounts"},
+				Scope:       &scope,
+			},
+		},
+	}
+	allowedUsers  []string
+	allowedGroups = []string{
+		"system:serviceaccounts:openshift-backplane-srep",
+	}
+	allowedServiceAccounts = []string{
+		"builder",
+		"default",
+		"deployer",
+	}
+	exceptionNamespaces = []string{
+		"openshift-logging",
+		"openshift-user-workload-monitoring",
+		"openshift-operators",
+	}
+)
+
+type serviceAccountWebhook struct {
+	s runtime.Scheme
+}
+
+// NewWebhook creates the new webhook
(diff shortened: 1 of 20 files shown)


feichashao:
Had a discussion with @bmeng , the main changes are `serviceaccount.go` and the corresponding tests.  Also added a "namespace" parameters to the `CreateHTTPRequest` in `testutils.go`.
The other changes are auto generated.


The test cases match the expectation of spiking in OSD-12731.

pkg/webhooks/serviceaccount/serviceaccount.go
```diff
@@ -0,0 +1,222 @@
...
+			Operations: []admissionregv1.OperationType{"DELETE"},
+			Rule: admissionregv1.Rule{
+				APIGroups:   []string{""},
+				APIVersions: []string{"v1"},
+				Resources:   []string{"serviceaccounts"},
+				Scope:       &scope,
+			},
+		},
+	}
+	allowedUsers  []string
```
wanghaoran1988, line 40: is this intent to be empty?

pkg/webhooks/serviceaccount/serviceaccount.go
```diff
@@ -0,0 +1,222 @@
...
+	allowedGroups = []string{
+		"system:serviceaccounts:openshift-backplane-srep",
+	}
+	allowedServiceAccounts = []string{
```
wanghaoran1988, line 44: Should we allow them to delete these?

pkg/webhooks/serviceaccount/serviceaccount.go
```diff
@@ -0,0 +1,222 @@
...
+		"builder",
+		"default",
+		"deployer",
+	}
+	protectedNamespaces = regexp.MustCompile(managedNamespaces)
+	exceptionNamespaces = []string{
```
wanghaoran1988, line 50: Could we put those namespaces somewhere else, so it can be used by other webhooks?

pkg/webhooks/serviceaccount/serviceaccount.go
bmeng: It is preserved for future if we would like to add users to have the ability to delete.

pkg/webhooks/serviceaccount/serviceaccount.go
bmeng: During my testing, those serviceaccounts can be recovered automatically after deletion, and no function impacted after that.

pkg/webhooks/serviceaccount/serviceaccount.go
bmeng: Do you mean we can abstract the variable in a different file, so all the webhooks can call them?
What I am thinking is if we control the exception list within the webhook, it would be easy for us to tweak this particular webhook.
It would be easy for testing as well IMO.

pkg/webhooks/serviceaccount/serviceaccount.go
```diff
@@ -0,0 +1,222 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"net/http"
+	"os"
+	"regexp"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	admissionctl "sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
+	"github.com/openshift/managed-cluster-validating-webhooks/pkg/webhooks/utils"
```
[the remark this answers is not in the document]
bmeng: updated

bmeng:
/hold 

pkg/webhooks/serviceaccount/serviceaccount.go
```diff
@@ -0,0 +1,223 @@
...
+
+	return false
+}
+
+// isProtectedNamespace checks if the request is going to operate on the serviceaccount in the
+// protected namespace list
+func isProtectedNamespace(request admissionctl.Request) bool {
+	ns := request.Namespace
+
+	if protectedNamespaces.Match([]byte(ns)) && !utils.SliceContains(ns, exceptionNamespaces) {
```
NautiluX, line 147: Is the list of namespaces we want to protect by this webhook different from https://github.com/openshift/managed-cluster-validating-webhooks/blob/408009d2a597965df8ed020413aab8d9ec004793/pkg/config/namespaces.go#L11 ?

NautiluX (commented):
We have a list of protected namespaces we use in other areas in this code. Is this list different from the list of namespaces for this webhook?
I think it's generated from https://github.com/openshift/managed-cluster-config/blob/master/resources/managed/all-osd-resources.yaml

pkg/webhooks/serviceaccount/serviceaccount.go
bmeng: thanks! updated to use the privilegedNamespace list in `config`

NautiluX (commented):
/lgtm

PR #230, opened 2023-06-27
[OSD-16225] : Add clusterrolebinding webhooks to prevent deletion of clusterrolebings in openshift and kube-system namespace

samanthajayasinghe:
Add the webhook for critical clusterrolebinding deletion prevention in following NS 

- openshift-*
- kube-system

### DIFF
--- a/pkg/webhooks/clusterrolebinding/clusterrolebinding.go
+++ b/pkg/webhooks/clusterrolebinding/clusterrolebinding.go
@@ -1,0 +1,200 @@
+package clusterrolebinding
+
+import (
+	"fmt"
+	"net/http"
+	"os"
+	"regexp"
+
+	"github.com/openshift/managed-cluster-validating-webhooks/pkg/webhooks/utils"
+	admissionv1 "k8s.io/api/admission/v1"
+	admissionregv1 "k8s.io/api/admissionregistration/v1"
+	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	admissionctl "sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+const (
+	WebhookName       string = "clusterrolebindings-validation"
+	docString         string = `Managed OpenShift Customers may not delete the cluster role bindings under the managed namespaces: %s`
+	managedNamespaces string = `(^openshift-.*|kube-system)`
+)
+
+var (
+	timeout int32 = 2
+	log           = logf.Log.WithName(WebhookName)
+	scope         = admissionregv1.ClusterScope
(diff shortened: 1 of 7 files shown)


pkg/webhooks/clusterrolebinding/clusterrolebinding.go
```diff
@@ -0,0 +1,200 @@
...
+	return fmt.Sprintf(docString, managedNamespaces)
+}
+
+// SyncSetLabelSelector returns the label selector to use in the SyncSet.
+// Return utils.DefaultLabelSelector() to stick with the default
+func (s *ClusterRoleBindingWebHook) SyncSetLabelSelector() metav1.LabelSelector {
+	return utils.DefaultLabelSelector()
+}
+
+func (s *ClusterRoleBindingWebHook) HypershiftEnabled() bool { return true }
```
bmeng, line 200: I think you will need to run `make package` to generate package manifests for hypershift

pkg/webhooks/clusterrolebinding/clusterrolebinding.go
```diff
@@ -0,0 +1,200 @@
...
+				APIVersions: []string{"v1"},
+				Resources:   []string{"clusterrolebindings"},
+				Scope:       &scope,
+			},
+		},
+	}
+
+	protectedNamespaces = regexp.MustCompile(managedNamespaces)
+
+	exceptionNamespaces = []string{
```
bmeng, line 44: Do we still need to validate the namespace if it is ClusterScope?

bmeng:
/hold 

pkg/webhooks/clusterrolebinding/clusterrolebinding.go
samanthajayasinghe: yeh, `clusterrolebindings` links to the subject  `ServiceAccount` which always comes through a namespace. 
basically, this var allows modifying `clusterrolebindings` which links `exception-Namespaces`

pkg/webhooks/clusterrolebinding/clusterrolebinding.go
samanthajayasinghe: Thanks @bmeng , generated the mainfests for HS

bmeng: @mrbarge @ravitri  Any of you know if we have criteria for adding webhook to hypershift clusters?

bmeng:
@samanthajayasinghe could you run the `make docs` to generate the document for the new webhook?

pkg/webhooks/clusterrolebinding/clusterrolebinding.go
mrbarge: Nothing documented. I think it's a case-by-case basis of asking the question:
- could a cluster owner even do this action from their hosted cluster? (if no: then not needed)
- if a cluster owner does this action from their hosted cluster, would it potentially bring the cluster into an unsupportable state? (if yes, probably needed)

In this case I think if we want this protection for OSD/ROSA it makes sense to also protect it in hosted clusters.

PR #231, opened 2023-06-27
Enable FIPS and switch to golang-1.19

mrbarge:
### What type of PR is this?
Feature

### What this PR does / why we need it?
- Switch to golang 1.19 for building MCVW and the selectorsyncset resources.
- Ensure that MCVW runs with FIPS enabled so that it is consistent with other OSD operators and workloads.

### Pre-checks (if applicable):
- [X] Tested latest changes against a cluster
- [X] Ran `make generate` command locally to validate code changes
- [ ] Included documentation changes with PR


mjlshen (approved):
/lgtm

PR #235, opened 2023-07-11
Allow kube and system users to delete service accounts

cblecker:
Found a bug where the backplane-cluster-admin user and system users like kube:admin could not delete service accounts


mjlshen (approved):
/lgtm

PR #237, opened 2023-07-12
OSD-16225: Allow kube and system users to delete cluster role bindings

samanthajayasinghe:
**Type**
Bug 

**Why ?**
SREP users and kube admin can't delete cluster role bindings in `openshift-*|kube-system` ns


mjlshen (commented):
/lgtm

PR #238, opened 2023-07-21
OSD-17575: Update webhook docs for ICSPs

cblecker:
This statement isn't exactly right. Updating it to be correct and slightly more generic.

### DIFF
--- a/pkg/webhooks/imagecontentpolicies/imagecontentpolicies.go
+++ b/pkg/webhooks/imagecontentpolicies/imagecontentpolicies.go
@@ -17,7 +17,7 @@ import (
 
 const (
 	WebhookName = "imagecontentpolicies-validation"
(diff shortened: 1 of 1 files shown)


pkg/webhooks/imagecontentpolicies/imagecontentpolicies.go
```diff
@@ -17,7 +17,7 @@ import (
 
 const (
 	WebhookName = "imagecontentpolicies-validation"
-	WebhookDoc  = "Managed OpenShift customers may not create ImageContentSourcePolicy, ImageDigestMirrorSet, or ImageTagMirrorSet resources that configure mirrors for the entirety of quay.io, registry.redhat.io, nor registry.access.redhat.com. If needed, specific repositories can have mirrors configured, such as quay.io/example."
+	WebhookDoc  = "Managed OpenShift customers may not create ImageContentSourcePolicy, ImageDigestMirrorSet, or ImageTagMirrorSet resources that configure mirrors that would conflict with system registries (e.g. quay.io, registry.redhat.io, registry.access.redhat.com, etc). For more details, see https://docs.openshift.com/"
```
mjlshen, line 20: I really like the reworded message, but I'm not sure what a customer would do with https://docs.openshift.com

mjlshen: documentation will exist in the future
/lgtm

PR #242, opened 2023-08-29
OSD-16469 - Add ingress controller webhook

hectorakemp:
Reverts openshift/managed-cluster-validating-webhooks#239

@fahlmant - since we are close to rolling out managed ingress v2 - it's probably safe to re-enable these changes. 

I will also put them behind the `legacy-ingress-support` flag to make the feature as safe as possible to in production.


hectorakemp:
Test here: https://gitlab.cee.redhat.com/-/snippets/7126

PR #243, opened 2023-09-13
Added openshift-gitops project as an exception per request in OSD-17844

madhusudanupadhyay:
Change made per: https://issues.redhat.com/browse/OSD-17844 to add openshift-gitops project as exception


mjlshen (commented):
/lgtm

PR #244, opened 2023-09-21
(SDE-3450) - Add gaurdrails to RH managed namespaces against NetworkPolicies

dakotalongRH:
Create a webhook that will prevent user from creating NetworkPolicies in Red Hat managed namespaces.


pkg/webhooks/networkpolicies/networkpolicies.go
```diff
@@ -0,0 +1,195 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return networkPolicy, nil
+}
+
+// GetURI implements Webhook interface
+func (s *networkpoliciesruleWebhook) GetURI() string {
+	return "/" + WebhookName
+}
```
lnguyen1401, line 147: I think it's missing a validate function like this 

 `
func (s *networkpoliciesruleWebhook) Validate(request admissionctl.Request) bool {
	valid := true
	valid = valid && (request.UserInfo.Username != "")
	valid = valid && (request.Kind.Kind == "NetworkPolicy")

	return valid
}
`

dakotalongRH: Good catch - added the function

lnguyen1401 (commented):
Missing registering webhook https://github.com/openshift/managed-cluster-validating-webhooks/tree/07bf070a0dcc1e564d5ff015bbf198ad13d6a79f#adding-new-webhooks

pkg/webhooks/networkpolicies/networkpolicies.go
```diff
@@ -0,0 +1,204 @@
...
+func (s *networkpoliciesruleWebhook) Authorized(request admissionctl.Request) admissionctl.Response {
+	//Implement authorized next
+	return s.authorized(request)
+}
+
+func (s *networkpoliciesruleWebhook) authorized(request admissionctl.Request) admissionctl.Response {
+	var ret admissionctl.Response
+
+	np, err := s.renderNetworkPolicy(request)
+
```
tnierman, line 79: nit:
```suggestion
```

pkg/webhooks/networkpolicies/networkpolicies.go
```diff
@@ -0,0 +1,204 @@
...
+	if err != nil {
+		log.Error(err, "Could not render a NetworkPolicy from the incoming request")
+		return admissionctl.Errored(http.StatusBadRequest, err)
+	}
+
+	if hookconfig.IsPrivilegedNamespace(np.GetNamespace()) {
+		log.Info(fmt.Sprintf("%s operation detected on managed namespace: %s", request.Operation, np.GetNamespace()))
+		if isAllowedUser(request) {
+			ret = admissionctl.Allowed(fmt.Sprintf("User can do operations on NetworkPolicies"))
```
tnierman, line 88: I think we should print the username here:
```suggestion
			ret = admissionctl.Allowed(fmt.Sprintf("User '%s' in group(s) '%s' can do operations on NetworkPolicies", request.UserInfo.Username, strings.Join(request.UserInfo.Groups, ", ")))
```
(untested)

Otherwise, we should remove the `fmt.Sprintf()` call, since we aren't formatting anything here

pkg/webhooks/networkpolicies/networkpolicies.go
```diff
@@ -0,0 +1,204 @@
...
+			ret.UID = request.AdmissionRequest.UID
+			return ret
+		}
+		for _, group := range request.UserInfo.Groups {
+			if privilegedServiceAccountGroupsRe.Match([]byte(group)) {
+				ret = admissionctl.Allowed("Privileged service accounts do operations on NetworkPolicies")
```
tnierman, line 94: Same as above, I think we should print the group that granted access for debugging purposes:
```suggestion
				ret = admissionctl.Allowed("Privileged service accounts in group '%s' can operate on NetworkPolicies", group)
```

pkg/webhooks/networkpolicies/networkpolicies.go
```diff
@@ -0,0 +1,204 @@
...
+				ret.UID = request.AdmissionRequest.UID
+				return ret
+			}
+		}
+
+		ret = admissionctl.Denied(fmt.Sprintf("Prevented from accessing Red Mat managed resources. This is in an effort to prevent harmful actions that may cause unintended consequences or affect the stability of the cluster. If you have any questions about this, please reach out to Red Hat support at https://access.redhat.com/support"))
```
tnierman, line 100: We can remove the `fmt.Sprintf()` call since we aren't formatting anything:
```suggestion
		ret = admissionctl.Denied("Prevented from accessing Red Mat managed resources. This is in an effort to prevent harmful actions that may cause unintended consequences or affect the stability of the cluster. If you have any questions about this, please reach out to Red Hat support at https://access.redhat.com/support")
```

pkg/webhooks/networkpolicies/networkpolicies.go
```diff
@@ -0,0 +1,204 @@
...
+			ret.UID = request.AdmissionRequest.UID
+			return ret
+		}
+		for _, group := range request.UserInfo.Groups {
+			if privilegedServiceAccountGroupsRe.Match([]byte(group)) {
+				ret = admissionctl.Allowed("Privileged service accounts do operations on NetworkPolicies")
+				ret.UID = request.AdmissionRequest.UID
+				return ret
+			}
+		}
```
tnierman, line 98: I think my personal preference would be to move this entire loop into the `isAllowedUser()` function, since it serves the same purpose. Not necessarily a blocker though

PR #246, opened 2023-10-24
OSD-18242 block customer modifying ingress config

yithian:
This webhook will prevent non-privileged serviceaccounts from creating, updating or deleting ingresses from config.openshift.io

Modifications to the cluster ingress config can degrade the authentication, console, and ingress operators

### DIFF
--- a/README.md
+++ b/README.md
@@ -318,6 +318,7 @@ At this point, we have:
 * regular-user-validation-osd
 * scc-validation
 * techpreviewnoupgrade-validation
+* ingress-config-validation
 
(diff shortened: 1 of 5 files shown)


yithian:
I've deployed these changes to a stage cluster and confirmed that a user with cluster-admin privileges cannot modify the ingress config:

```
$ oc edit ingress.config cluster
error: ingresses.config.openshift.io "cluster" could not be patched: admission webhook "ingress-config-validation.managed.openshift.io" denied the request: Only privileged service accounts may access
You can run `oc replace -f /tmp/oc-edit-1762185798.yaml` to try this update again.
$ oc whoami
gram
$ oc get groups
NAME               USERS
cluster-admins     gram
dedicated-admins   
$ 
```

mjlshen (approved):
/label tide/merge-method-squash
/lgtm

PR #247, opened 2023-11-01
Deploy to non-request-serving nodes

AlexVulaj:
Resolves [OSD-18990](https://issues.redhat.com//browse/OSD-18990)


tnierman:
I must know

/trick-or-treat

PR #253, opened 2023-11-09
change the log for prometheus webhook less verbose

bmeng:
The request body will be pretty large for the prometheus rule resource. Remove the request body from the log.


mjlshen (approved):
/lgtm

PR #255, opened 2023-11-26
Propose deprecating this repo and moving to ValidatingAdmissionPolicy

mjlshen:
If there's agreement amongst approvers in this repo, the next step for this design will be an ADR.


mjlshen:
/hold

Would like to get positive feedback from at least 2 other approvers

PR #257, opened 2023-12-06
[WIP] Allow ocm agent operator to skip NP webhook

hectorakemp:


pkg/webhooks/networkpolicies/networkpolicies.go
```diff
@@ -26,7 +26,7 @@ const (
 
 var (
 	timeout                          int32 = 2
-	allowedUsers                           = []string{"system:admin", "backplane-cluster-admin"}
+	allowedUsers                           = []string{"system:admin", "backplane-cluster-admin", "system:serviceaccounts:openshift-ocm-agent-operator:ocm-agent-operator"}
 	sreAdminGroups                         = []string{"system:serviceaccounts:openshift-backplane-srep"}
 	privilegedServiceAccountGroupsRe       = regexp.MustCompile(privilegedServiceAccountGroups)
```
bmeng, line 31: The `system:serviceaccounts:openshift-ocm-agent-operator:ocm-agent-operator` should be part of this already? So might be a bug?

hectorakemp: It must be a bug :thinking:
Probably best to look into why exactly this isn't working in the first place cc @dakotalongRH

hectorakemp:
Closing this, because there is actually no bug with OCM service accounts. Original ticket refers to cluster admin not having access to edit network policies. All new tests actually pass with the original code.

PR #259, opened 2023-12-08
Restrict clusterversion create/patch/update actions

mjlshen:
Previously, any authenticated "system:\*" user could modify
clusterversion objects. This unfortunately would include any
"system:serviceaccount:\*" that exists on the cluster. This PR restricts
it to the only two known serviceaccounts that should modify
clusterversion objects in OSD/ROSA: MUO and CVO

[OSD-19683](https://issues.redhat.com//browse/OSD-19683)


mjlshen:
Makes more sense after #258 merges

Makefile
mjlshen, line 1: I made these changes because I was following the README to test on stage, but if I ran `make build-base` locally it predictably built ARM images which don't run on stage clusters :D

pkg/webhooks/regularuser/common/regularuser.go
```diff
@@ -211,36 +216,55 @@ func (s *RegularuserWebhook) authorized(request admissionctl.Request) admissionc
...
-		ret.UID = request.AdmissionRequest.UID
-		return ret
-	}
-	if isMustGatherAuthorized(request) {
-		ret = admissionctl.Allowed("Management of MustGather CR is authorized")
-		ret.UID = request.AdmissionRequest.UID
-		return ret
+
+	switch {
+	case request.Kind.Kind == mustGatherKind && request.Kind.Group == mustGatherGroup:
```
tnierman, line 227: Could we encapsulate this in a function to improve readability?
```suggestion
	case isMustGatherReq(request):
```

pkg/webhooks/regularuser/common/regularuser.go
```diff
@@ -263,28 +287,46 @@ func (s *RegularuserWebhook) authorized(request admissionctl.Request) admissionc
...
 	return (slices.Contains(request.UserInfo.Groups, "cluster-admins") ||
 		slices.Contains(request.UserInfo.Groups, "dedicated-admins")) &&
-		request.Kind.Kind == netNamespaceKind &&
-		request.Kind.Group == netNamespaceGroup &&
 		isNetNamespaceValid(s, request)
 }
 
+// isClusterVersionAuthorized only allows specific K8s serviceaccounts to modify ClusterVersion resources
+func isClusterVersionAuthorized(request admissionctl.Request) bool {
+	log.Info("analyzing clusterversion", "user", request.UserInfo.Username, "group", request.UserInfo.Groups)
```
tnierman, line 308:
```suggestion
```
Is this debugging output? I'd suggest we remove this to remain consistent with the other `is*Authorized()` functions

pkg/webhooks/regularuser/common/regularuser.go
mjlshen: I'm down, what do you think about something like:

```go
func requestMatchesGroupKind(req admissionctl.Request, kind, group string) bool
```

So we don't have to make a bunch of `is${KIND}Req` functions?

pkg/webhooks/regularuser/common/regularuser.go
mjlshen: Ah, I thought it was useful to keep, but happy to remove it for consistency

tnierman: We could always do the inverse as well, and add more logging :slightly_smiling_face:

mjlshen: I just took it out, not too attached to it

PR #262, opened 2024-01-04
OSD-19897 - latest config generation from MCC

theautoroboto:
This update was generated after updating managed-cluster-config.

`addons-namespaces.ConfigMap.yaml` is no longer being used and is being removed.  There is no corresponding `addons-namespaces.ConfigMap.yaml` in MCC 
https://github.com/openshift/managed-cluster-config/tree/master/deploy/osd-managed-resources


mjlshen (approved):
/lgtm

PR #264, opened 2024-01-10
[OSD-16432] Adds a validating webhook to block unpriviledged modifications of protected custom resource definitions

T0MASD:
Adds a validating webhook to block unpriviledged modifications of protected custom resource definitions while allowing modifications of unprotected ones.

### DIFF
--- a/pkg/webhooks/customresourcedefinitions/customresourcedefinitions.go
+++ b/pkg/webhooks/customresourcedefinitions/customresourcedefinitions.go
@@ -10,6 +10,7 @@ import (
 	"github.com/openshift/managed-cluster-validating-webhooks/pkg/webhooks/utils"
 
 	admissionregv1 "k8s.io/api/admissionregistration/v1"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
@@ -48,12 +49,6 @@ var (
 // customresourcedefinitionsruleWebhook validates a customresourcedefinition change
 type customresourcedefinitionsruleWebhook struct {
 	s runtime.Scheme
-}
-
-// We just need a runtime object to get the labels
-type customResourceDefinition struct {
-	runtime.Object
-	metav1.ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
 }
 
 // NewWebhook creates the new webhook
@@ -120,12 +115,12 @@ func isAllowedUser(request admissionctl.Request) bool {
 	return false
 }
 
-func (s *customresourcedefinitionsruleWebhook) renderCustomResourceDefinition(req admissionctl.Request) (*customResourceDefinition, error) {
+func (s *customresourcedefinitionsruleWebhook) renderCustomResourceDefinition(req admissionctl.Request) (*apiextensionsv1.CustomResourceDefinition, error) {
 	decoder, err := admissionctl.NewDecoder(&s.s)
 	if err != nil {
 		return nil, err
 	}
-	customResourceDefinition := &customResourceDefinition{}
+	customResourceDefinition := &apiextensionsv1.CustomResourceDefinition{}
 
 	if len(req.OldObject.Raw) > 0 {
 		err = decoder.DecodeRaw(req.OldObject, customResourceDefinition)
--- a/pkg/webhooks/customresourcedefinitions/customresourcedefinitions.go
+++ b/pkg/webhooks/customresourcedefinitions/customresourcedefinitions.go
@@ -80,7 +80,7 @@ func (s *customresourcedefinitionsruleWebhook) authorized(request admissionctl.R
 		return admissionctl.Errored(http.StatusBadRequest, err)
 	}
 
-	if utils.IsProtectedByLabel(np.GetLabels()) || utils.IsProtectedByResourceName(np.GetName()) {
+	if utils.IsProtectedByResourceName(np.GetName()) {
 		log.Info(fmt.Sprintf("%s operation detected on protected CustomResourceDefinition: %s", request.Operation, np.Name))
 		if isAllowedUser(request) {
 			ret = admissionctl.Allowed(fmt.Sprintf("User '%s' in group(s) '%s' can operate on CustomResourceDefinitions", request.UserInfo.Username, strings.Join(request.UserInfo.Groups, ", ")))
--- a/pkg/webhooks/customresourcedefinitions/customresourcedefinitions.go
+++ b/pkg/webhooks/customresourcedefinitions/customresourcedefinitions.go
@@ -17,15 +17,15 @@ import (
 )
 
 const (
-	WebhookName                    string = "customresourcedefinitions-validation"
-	docString                      string = `Managed OpenShift Customers may not change CustomResourceDefinitions managed by Red Hat.`
+	WebhookName string = "customresourcedefinitions-validation"
+	docString   string = `Managed OpenShift Customers may not change CustomResourceDefinitions managed by Red Hat.`
 )
 
 var (
 	timeout                          int32 = 2
 	allowedUsers                           = []string{"system:admin", "backplane-cluster-admin"}
 	sreAdminGroups                         = []string{"system:serviceaccounts:openshift-backplane-srep"}
-	privilegedServiceAccountGroupsRe       = regexp.MustCompile(privilegedServiceAccountGroups)
+	privilegedServiceAccountGroupsRe       = regexp.MustCompile(utils.PrivilegedServiceAccountGroups)
 	scope                                  = admissionregv1.ClusterScope
 	rules                                  = []admissionregv1.RuleWithOperations{
 		{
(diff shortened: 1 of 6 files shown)


pkg/webhooks/customresourcedefinitions/customresourcedefinitions.go
```diff
@@ -0,0 +1,203 @@
...
+func (s *customresourcedefinitionsruleWebhook) authorized(request admissionctl.Request) admissionctl.Response {
+	var ret admissionctl.Response
+
+	np, err := s.renderCustomResourceDefinition(request)
+	if err != nil {
+		log.Error(err, "Could not render a CustomResourceDefinition from the incoming request")
+		return admissionctl.Errored(http.StatusBadRequest, err)
+	}
+
+	if hookconfig.IsPrivilegedNamespace(np.GetNamespace()) {
```
mjlshen, line 84: Does this actually work? CRDs are not namespaced - `Namespaced` here refers to the corresponding CustomResources that are made

mjlshen (commented):
/hold

pkg/webhooks/customresourcedefinitions/customresourcedefinitions.go
T0MASD: from https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#create-a-customresourcedefinition CRDs can be Namespaced or Cluster

to get the list of namespaced CRD's on OSD:

oc get crd -o json | jq -r '.items[] | [.metadata.name, .spec.scope] | @tsv' | grep Namespaced

mjlshen: > The custom resource created from a CRD object can be either namespaced or cluster-scoped, as specified in the CRD's spec.scope field

but you can verify this in practice by trying to `oc get crd -n ${NS}` - it's not a namespaced object

T0MASD: ack, makes sense now. It appears the way to handle this would be:
1. create a list of protected crd names
2. check the incoming request crd .metadata.name in the list of protected crds
3. if true and allowed return admissionctl.Response(Allowed)
4. if true and not allowed return admissionctl.Response(Denied)
5. if false return admissionctl.Response(Allowed)

T0MASD: validating by label added in [aa86d0a](https://github.com/openshift/managed-cluster-validating-webhooks/pull/264/commits/aa86d0aa4b230d8687a947917ce79a52b587cfe4) let's use new thread to discuss further changes

pkg/webhooks/customresourcedefinitions/customresourcedefinitions.go
```diff
@@ -0,0 +1,202 @@
...
+	WebhookName                    string = "customresourcedefinitions-validation"
+	docString                      string = `Managed OpenShift Customers may not change CustomResourceDefinitions managed by Red Hat.`
+	privilegedServiceAccountGroups string = `^system:serviceaccounts:(kube.*|openshift.*|default|redhat.*|osde2e-[a-z0-9]{5})`
+)
+
+var (
+	timeout                          int32 = 2
+	allowedUsers                           = []string{"system:admin", "backplane-cluster-admin"}
+	sreAdminGroups                         = []string{"system:serviceaccounts:openshift-backplane-srep"}
+	privilegedServiceAccountGroupsRe       = regexp.MustCompile(privilegedServiceAccountGroups)
```
mjlshen, line 29: Could we use `utils.PrivilegedServiceAccountGroups` instead there was a recent finding that needed to update this regex string?

pkg/webhooks/customresourcedefinitions/customresourcedefinitions.go
```diff
@@ -0,0 +1,202 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	admissionctl "sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+const (
+	WebhookName                    string = "customresourcedefinitions-validation"
+	docString                      string = `Managed OpenShift Customers may not change CustomResourceDefinitions managed by Red Hat.`
+	privilegedServiceAccountGroups string = `^system:serviceaccounts:(kube.*|openshift.*|default|redhat.*|osde2e-[a-z0-9]{5})`
```
mjlshen, line 22:
```suggestion
```

There's a recently added `utils.PrivilegedServiceAccountGroups`

pkg/webhooks/customresourcedefinitions/customresourcedefinitions.go
```diff
@@ -0,0 +1,202 @@
...
+func (s *customresourcedefinitionsruleWebhook) authorized(request admissionctl.Request) admissionctl.Response {
+	var ret admissionctl.Response
+
+	np, err := s.renderCustomResourceDefinition(request)
+	if err != nil {
+		log.Error(err, "Could not render a CustomResourceDefinition from the incoming request")
+		return admissionctl.Errored(http.StatusBadRequest, err)
+	}
+
+	if utils.IsProtectedByLabel(np.GetLabels()) || utils.IsProtectedByResourceName(np.GetName()) {
```
mjlshen, line 83:
```suggestion
	if utils.IsProtectedByResourceName(np.GetName()) {
```

Let's scope this initial effort to only protecting by a whitelist, since we do not yet know if the label direction is where we want to go (would need collaboration with OCP).

pkg/webhooks/customresourcedefinitions/customresourcedefinitions.go
```diff
@@ -0,0 +1,202 @@
...
+// Authorized implements Webhook interface
+func (s *customresourcedefinitionsruleWebhook) Authorized(request admissionctl.Request) admissionctl.Response {
+	//Implement authorized next
+	return s.authorized(request)
+}
+
+func (s *customresourcedefinitionsruleWebhook) authorized(request admissionctl.Request) admissionctl.Response {
+	var ret admissionctl.Response
+
+	np, err := s.renderCustomResourceDefinition(request)
```
mjlshen, line 77:
```suggestion
	crd, err := s.renderCustomResourceDefinition(request)
```
nit, but the `np` variable name doesn't really make sense to me

pkg/webhooks/customresourcedefinitions/customresourcedefinitions.go
```diff
@@ -0,0 +1,202 @@
...
+	if err != nil {
+		return nil, err
+	}
+	customResourceDefinition := &customResourceDefinition{}
+
+	if len(req.OldObject.Raw) > 0 {
+		err = decoder.DecodeRaw(req.OldObject, customResourceDefinition)
+	} else {
+		err = decoder.Decode(req, customResourceDefinition)
+	}
```
mjlshen, line 135:
```suggestion
  err = decoder.Decode(req, customResourceDefinition)
```
Why do we prefer to decode OldObject instead of Object?

mjlshen (commented):
Thanks for working on this, my main feedback is around removing the label filtering logic for now. We should implement that feature separately as it's requires more collaboration.

pkg/webhooks/customresourcedefinitions/customresourcedefinitions.go
T0MASD: added in a973e244c24dcccbde17fd4d3078e746d410ecb6

pkg/webhooks/customresourcedefinitions/customresourcedefinitions.go
T0MASD: hm, I tried using the Object, however then normal user and backplane admin could not perform delete method. FWIW this logic is used in other validators
```
pkg/webhooks/namespace/namespace.go-134-        if len(req.OldObject.Raw) > 0 {
pkg/webhooks/namespace/namespace.go-160-        if len(req.OldObject.Raw) == 0 {
pkg/webhooks/pod/pod.go-102-    if len(req.OldObject.Raw) > 0 {
pkg/webhooks/prometheusrule/prometheusrule.go-167-      if len(req.OldObject.Raw) > 0 {
pkg/webhooks/networkpolicies/networkpolicies.go-131-    if len(req.OldObject.Raw) > 0 {
pkg/webhooks/customresourcedefinitions/customresourcedefinitions.go-130-        if len(req.OldObject.Raw) > 0 {
```

pkg/webhooks/customresourcedefinitions/customresourcedefinitions.go
```diff
@@ -0,0 +1,201 @@
...
+// customresourcedefinitionsruleWebhook validates a customresourcedefinition change
+type customresourcedefinitionsruleWebhook struct {
+	s runtime.Scheme
+}
+
+// We just need a runtime object to get the labels
+type customResourceDefinition struct {
+	runtime.Object
+	metav1.ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
+}
```
mjlshen, line 57: We shouldn't be defining our own type here and instead use the pre-defined type upstream within apiextensions, e.g. https://github.com/openshift/managed-cluster-validating-webhooks/blob/master/pkg/webhooks/namespace/namespace.go#L158

mjlshen: Here's the CRD type upstream https://github.com/kubernetes/apiextensions-apiserver/blob/master/pkg/apis/apiextensions/v1/types.go#L367C24-L379

pkg/webhooks/utils/utils.go
```diff
@@ -41,6 +42,26 @@ func DefaultLabelSelector() metav1.LabelSelector {
 	}
 }
 
+func IsProtectedByLabel(labels map[string]string) bool {
+	return labels["managed.openshift.io/protected"] == "true"
+}
```
mjlshen, line 47:
```suggestion
```

pkg/webhooks/customresourcedefinitions/customresourcedefinitions.go
```diff
@@ -0,0 +1,201 @@
...
+
+	return false
+}
+
+func (s *customresourcedefinitionsruleWebhook) renderCustomResourceDefinition(req admissionctl.Request) (*customResourceDefinition, error) {
+	decoder, err := admissionctl.NewDecoder(&s.s)
+	if err != nil {
+		return nil, err
+	}
+	customResourceDefinition := &customResourceDefinition{}
```
mjlshen, line 128:
```suggestion
	customResourceDefinition := &apiextensionsv1.CustomResourceDefinition{}
```
Once the import `apiextensionsv1` is added for https://github.com/kubernetes/apiextensions-apiserver/blob/master/pkg/apis/apiextensions/v1/types.go#L367

mjlshen (approved):
/unhold
/approve

PR #270, opened 2024-01-19
Networkpolicies: Fix UPDATE requests

NautiluX:
* UPDATE requests validated the OldObjects of the request
* OldObject is only used for update and delete requests, so when a new object exists it's always that which should be validated

Jira: [OSD-20308](https://issues.redhat.com//browse/OSD-20308)


mjlshen (approved):
/lgtm

PR #274, opened 2024-01-24
[OSD-20485] Removed osd/regularuser webhook

aliceh:
In this commit, I removed the osd/regularuser webhook in preparation for updating the node webhook.
This should not change the collective behavior of our webhooks, i.e. same users should be blocked from performing the same actions as before.

https://issues.redhat.com/browse/OSD-20485


mjlshen (approved):
/lgtm

PR #275, opened 2024-01-24
[OSD-20485]Updated node webhook to allow user to modify worker nodes

aliceh:
In this commit, I modified the node webhook to allow OSD and ROSA users to modify worker nodes.

https://issues.redhat.com/browse/OSD-20485


mjlshen (approved):
/lgtm

PR #276, opened 2024-01-24
[OSD-20485]Added an explicit dent for users trying to delete a worker node

aliceh:
In this commit, I added a deny policy for users deleting a worker node.


mjlshen (approved):
/lgtm

PR #280, opened 2024-02-01
OSD-20474 add a webhook to prevent OVN<->SDN

yithian:
prevent patching of network.config.openshift.io resources to change the networkType

### DIFF
--- a/pkg/webhooks/sdnmigration/sdnmigration.go
+++ b/pkg/webhooks/sdnmigration/sdnmigration.go
@@ -2,14 +2,11 @@ package sdnmigration
 
 import (
 	"net/http"
-	"os"
 	"regexp"
-	"sync"
 
 	configv1 "github.com/openshift/api/config/v1"
 	admissionv1 "k8s.io/api/admission/v1"
 	admissionregv1 "k8s.io/api/admissionregistration/v1"
-	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
@@ -45,7 +42,7 @@ var (
 )
 
 type NetworkConfigWebhook struct {
-	s  runtime.Scheme
+	s runtime.Scheme
 }
 
 // Authorized will determine if the request is allowed
--- a/pkg/webhooks/sdnmigration/sdnmigration_test.go
+++ b/pkg/webhooks/sdnmigration/sdnmigration_test.go
@@ -268,7 +268,7 @@ func TestSyncSetLabelSelector(t *testing.T) {
 func TestHypershiftEnabled(t *testing.T) {
 	enabled := NewWebhook().HypershiftEnabled()
 
-	if !enabled {
-		t.Error("TestHypershiftEnabled(): expected enabled")
+	if enabled {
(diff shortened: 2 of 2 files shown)


pkg/webhooks/sdnmigration/sdnmigration.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+)
+
+type NetworkConfigWebhook struct {
+	mu sync.Mutex
+	s  runtime.Scheme
+}
+
+// Authorized will determine if the request is allowed
+func (w *NetworkConfigWebhook) Authorized(request admissionctl.Request) (ret admissionctl.Response) {
```
mjlshen, line 53:
```suggestion
func (w *NetworkConfigWebhook) Authorized(request admissionctl.Request) (admissionctl.Response) {
```

pkg/webhooks/sdnmigration/sdnmigration.go
```diff
@@ -0,0 +1,176 @@
...
+	log.Info("achvtest authorized() running")
+	ret = admissionctl.Denied("Only privileged service accounts may access")
+	ret.UID = request.AdmissionRequest.UID
+
+	// allow if modified by an allowlist-ed service account
+	for _, group := range request.UserInfo.Groups {
+		if privilegedServiceAccountsRe.Match([]byte(group)) {
+			ret = admissionctl.Allowed("Privileged service accounts may access")
+			ret.UID = request.AdmissionRequest.UID
```
mjlshen, line 62:
```suggestion
			ret.UID = request.AdmissionRequest.UID
			return ret
```

pkg/webhooks/sdnmigration/sdnmigration.go
```diff
@@ -0,0 +1,170 @@
...
+			ret = admissionctl.Errored(http.StatusBadRequest, err)
+			ret.UID = request.AdmissionRequest.UID
+		}
+		if err := decoder.DecodeRaw(request.OldObject, oldObject); err != nil {
+			log.Error(err, "failed to render a Network from request.OldObject")
+			ret = admissionctl.Errored(http.StatusBadRequest, err)
+			ret.UID = request.AdmissionRequest.UID
+		}
+
+		if object.Spec.NetworkType != oldObject.Status.NetworkType {
```
mjlshen, line 87:
```suggestion
		if object.Spec.NetworkType != oldObject.Spec.NetworkType {
```
Is there a reason to not compare `.spec.networkType` here? (Mostly out of curiosity)

mjlshen (commented):
LGTM overall, it works - couple questions/nits

pkg/webhooks/sdnmigration/sdnmigration.go
yithian: I was taking a cue from [the configv1 docs](https://pkg.go.dev/github.com/openshift/api/config/v1#Network) where it suggests reading from `Status` instead of `Spec`

> 	// spec holds user settable values for configuration.
	// As a general rule, this SHOULD NOT be read directly. Instead, you should
	// consume the NetworkStatus, as it indicates the currently deployed configuration.
	// Currently, most spec fields are immutable after installation. Please view the individual ones for further details on each.

mjlshen: Awesome, thanks for the explanation!

mjlshen (approved):
/lgtm

PR #281, opened 2024-02-01
Allow generation of MutatingWebhookConfigurations & add HCP-friendly MutatingWebhook for LoadBalancer-type Services

abyrne55:
This PR addresses [OSD-20193](https://issues.redhat.com/browse/OSD-20193), which calls for a MutatingWebhook to be installed across our fleet that ensures all Service-backed ELBs are marked with the resource tag required by managed policies. Currently, requests from HCP users attempting to create Services of type `LoadBalancer` will be rejected by the EC2 API unless the Service contains the annotation `service.beta.kubernetes.io/aws-load-balancer-additional-resource-tags: red-hat-managed=true`. The MutatingWebhook implemented by this PR adds that annotation to any CREATE or UPDATE requests touching a Service object. If the Service already contains an annotation with the key `service.beta.kubernetes.io/aws-load-balancer-additional-resource-tags`, the webhook ensures the annotation's value contains `red-hat-managed=true` while preserving any other user-defined tags (unless the user tries to do something cheeky like `red-hat-managed=false`).

Note: while managed policies currently only apply to HCPs, this webhook is designed to be launched across both HCPs and ROSA Classic clusters. 

**Design Details**
Despite the K8s SDK types for both kinds of Webhooks being quite similar, they do not share any interfaces, so it's very difficult/ugly to refactor the existing ValidatingWebhook-oriented code to also accept MutatingWebhooks using generics/type parameters. Instead, this PR offers a side-by-side implementation where each type of Webhooks is processed by its own chain of functions. While this is simpler, easier to implement, and less prone to runtime panics than a generic implementation, it is also less maintainable due to heavily-duplicated code. IMO, this is an acceptable tradeoff because [this repo is soon to be deprecated](https://github.com/openshift/managed-cluster-validating-webhooks/blob/master/designs/validating_admission_policy.md).

**How to Test this PR**
Set up fresh OSD/ROSA Classic and HCP clusters in stage, then do the following for each:
1. Clone/checkout this PR, and run `make test`
2. Follow the instructions in this repo's README (or [this HCP-specific doc](https://github.com/openshift/managed-cluster-validating-webhooks/blob/master/docs/hypershift.md) if installing into a HostedCluster) for installing the new MutatingWebhook (named `service-mutation`) into your cluster
3. Create a LoadBalancer-backed service without the necessary annotation/tag value, e.g., by saving the example service below and running `oc apply -f examplesvc.yaml`
```yaml
apiVersion: v1
kind: Service
metadata:
  name: testservice-20193
  annotations: {"service.beta.kubernetes.io/aws-load-balancer-additional-resource-tags" : "Org=Engineering,Dept=SRE"}
spec:
  ports:
  - name: hello-node
    port: 80
  loadBalancerIP:
  loadBalancerSourceRanges: 
  - 10.0.0.0/8
  - 192.168.0.0/16
  type: LoadBalancer 
  selector:
    name: hello-node 
```
4. In the AWS console for the account hosting your cluster's worker nodes, ensure the newly-created ELB contains the `red-hat-managed=true` tag as well as any other resource tags you specified.
5. Attempt to remove the tag, e.g., by running `oc annotate service testservice-20193 service.beta.kubernetes.io/aws-load-balancer-additional-resource-tags-`
6. Back in the AWS console, ensure the `red-hat-managed=true` tag survived

/label tide/merge-method-squash
/kind feature


build/resources.go
mjlshen, line 1: We need to add this to a CI check :( but have you run `make package`?

mjlshen (approved):
/lgtm

PR #283, opened 2024-02-08
Better enable internal testing of SDN --> OVN

mjlshen:
Add a "red-hat-internal-testing": "true" override to disable the effects of this webhook.

The unit tests weren't working as intended because the decoder interacts with the `Raw` object inside of each admission requests's Object/OldObject so I fixed those up as well.

[OSD-20474](https://issues.redhat.com//browse/OSD-20474)

### DIFF
--- a/pkg/webhooks/sdnmigration/sdnmigration.go
+++ b/pkg/webhooks/sdnmigration/sdnmigration.go
@@ -18,14 +18,13 @@ import (
 const (
 	WebhookName               string = "sdn-migration-validation"
 	privilegedServiceAccounts string = `^system:serviceaccounts:(kube.*|openshift.*|default|redhat.*|osde2e-[a-z0-9]{5})`
-	privilegedUsers           string = `system:admin`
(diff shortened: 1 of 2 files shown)


pkg/webhooks/sdnmigration/sdnmigration.go
```diff
@@ -18,14 +18,13 @@ import (
 const (
 	WebhookName               string = "sdn-migration-validation"
 	privilegedServiceAccounts string = `^system:serviceaccounts:(kube.*|openshift.*|default|redhat.*|osde2e-[a-z0-9]{5})`
-	privilegedUsers           string = `system:admin`
 	docString                 string = `Managed OpenShift customers may not modify the network config type because it can can degrade cluster operators and can interfere with OpenShift SRE monitoring.`
+	overrideAnnotation        string = "red-hat-internal-testing"
```
tnierman, line 22: Kinda nit-picky, but thoughts on updating the annotation to
```suggestion
	overrideAnnotation        string = "unsupported-red-hat-internal-testing"
```
to emphasize that this annotation should definitely not be used if you don't know what you're doing?

PR #284, opened 2024-02-08
Remove ServiceAccount from packaged deployment

abyrne55:
This PR addresses bug [OSD-20871](https://issues.redhat.com/browse/OSD-20871), in which MCVW's HCP package failed to deploy on management clusters due to an extraneous reference to a service account. None of our HyperShift-enabled webhooks need access to the cluster's API, so this PR removes that reference.


mjlshen (approved):
/lgtm

PR #279, opened 2024-01-25
Add feature flag to enable certain clusters to drain worker nodes

mjlshen:
OSD-20485

PR #285, opened 2024-02-15
OSD-20487 Revert "Add feature flag to enable certain clusters to drain worker nodes (#279)"

yithian:
This reverts commit f6216b69a667f07588e3dad56739e4aa5950f361.

Step 1 for enabling all customers to drain/cordon/uncordon worker nodes


iamkirkbater:
/lgtm

(I don't think I have permissions for that to work).

We should consider using this as a reference in future work for pulling from a configmap at webhook startup.

PR #261, opened 2023-12-28
[OSD-12399] Block Modify/Edit of kubeletConfig using webhook

itsmitul9:
- Added kubeleconfig.go under webhooks/kubeletconfig 
- Added add_kubeletconfig.go under webhooks 
- Added kubeletconfig-validation webhook description in webhook.json and webhook-short.json

PR #290, opened 2024-03-18
[OSD-12399] Block Modify/Edit of kubeletConfig using webhook

itsmitul9:


mjlshen:
From #261

/hold

The card ends with:

the webhook method would block ${REDACTED} use cases, even though it's more preferable. the splunk+OHSS ticket would create more toil, so maybe OCM agent to sent the SL

Did something change to allow us to use validating webhooks to block this?

PR #291, opened 2024-04-05
Allow osde2e to modify SCCs

mjlshen:
osde2e runs a conformance test which needs to be able to modify SCCs https://github.com/openshift/kubernetes/blob/e994e5dbd69833088514616351b0aa997e4ed79d/openshift-hack/e2e/namespace.go#L64-L71

This PR accomplishes this by adding a regexp for the osde2e group as a privileged group that can modify SCCs at will.

[OSD-22059](https://issues.redhat.com/browse/OSD-22059)

### DIFF
--- a/pkg/webhooks/scc/scc.go
+++ b/pkg/webhooks/scc/scc.go
@@ -3,14 +3,13 @@ package scc
 import (
 	"fmt"
 	"net/http"
-	"os"
+	"regexp"
 	"slices"
 
 	securityv1 "github.com/openshift/api/security/v1"
 	"github.com/openshift/managed-cluster-validating-webhooks/pkg/webhooks/utils"
 	admissionv1 "k8s.io/api/admission/v1"
 	admissionregv1 "k8s.io/api/admissionregistration/v1"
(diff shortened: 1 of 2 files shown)


pkg/webhooks/scc/scc.go
```diff
@@ -59,25 +58,13 @@ var (
...
-	if err != nil {
-		log.Error(err, "Fail adding admissionsv1 scheme to SCCWebHook")
-		os.Exit(1)
-	}
-	err = corev1.AddToScheme(scheme)
-	if err != nil {
-		log.Error(err, "Fail adding corev1 scheme to SCCWebHook")
-		os.Exit(1)
-	}
-
```
tnierman, line 78: It looks like we're no longer adding `admissionv1` and `corev1` to our scheme. Is that because they're no longer needed in the webhook logic? Or does `runtime.NewScheme()` somehow know to add those by default?

tnierman: TIL :slightly_smiling_face:

PR #297, opened 2024-05-03
Only authorize cluster admins for mustgather requests

AlexVulaj:
Adds a webhook check to make sure that users making requests for mustgather CRDs are cluster admins.


pkg/webhooks/regularuser/common/regularuser.go
```diff
@@ -287,7 +288,7 @@ func (s *RegularuserWebhook) authorized(request admissionctl.Request) admissionc
 
 // isMustGatherAuthorized check if request is authorized for MustGather CR
 func isMustGatherAuthorized(request admissionctl.Request) bool {
-	return slices.Contains(request.UserInfo.Groups, ceeGroup)
```
mjlshen, line 290: I don't think ceeGroup should be removed here. We want CEE to be able to use must-gather-operator

mjlshen (changes requested):
Could you also link a card to this PR?

pkg/webhooks/regularuser/common/regularuser.go
AlexVulaj: I just added the ticket to this PR - there should be more details there.

AlexVulaj:
/hold

discussing internally

AlexVulaj:
We've decided to not move forward with this change.

PR #299, opened 2024-05-13
Remove osde2e exception from modifying SCCs

cblecker:
Partial revert of #291

ref [OSD-21840](https://issues.redhat.com//browse/OSD-21840)


mjlshen (approved):
/approve
/hold

Holding until CICD is aware of this change. Once that's the case, feel free to unhold whenever

PR #305, opened 2024-05-28
Add Mutating Webhook PodImageSpec

lnguyen1401:
Creating a mutating webhook that allow for image-registry to be disabled 
https://issues.redhat.com/browse/SDE-4088


pkg/webhooks/podimagespec/podimagespec.go
```diff
@@ -0,0 +1,281 @@
...
+
+	err = s.kubeClient.Get(ctx, client.ObjectKey{Name: "cluster"}, &s.configV1)
+	if err != nil {
+		return false, fmt.Errorf("failed to get image registry config: %v", err)
+	}
+
+	if s.configV1.Spec.ManagementState == operatorv1.Managed || s.configV1.Spec.ManagementState == operatorv1.Unmanaged {
+		return true, nil
+	} else if s.configV1.Spec.ManagementState == operatorv1.Removed {
+		return false, nil
```
rendhalver, line 88: I don't understand why we are returning false if the image-registry has been removed. Shouldn't we be returning true here like we are doing when it's unmanaged?

pkg/webhooks/podimagespec/podimagespec.go
```diff
@@ -0,0 +1,281 @@
...
+	decoder, err := admissionctl.NewDecoder(s.s)
+	if err != nil {
+		return nil, err
+	}
+	pod := &corev1.Pod{}
+	if len(request.OldObject.Raw) > 0 {
+		err = decoder.DecodeRaw(request.OldObject, pod)
+	} else {
+		err = decoder.DecodeRaw(request.Object, pod)
+	}
```
abyrne55, line 169:
```suggestion
	err = decoder.Decode(req, pod)
```
I made the mistake of doing this `OldObject` checking in my original version of this webhook and have since patched it. See #298 for details, but tl;dr is don't bother with `OldObject` unless you want really difficult-to-reproduce bugs. You might also want to see the ways I updated the unit tests in that PR

pkg/webhooks/podimagespec/podimagespec.go
```diff
@@ -0,0 +1,281 @@
...
+func (s *PodImageSpecWebhook) CheckImageRegistryStatus(ctx context.Context) (bool, error) {
+	var err error
+	if s.kubeClient == nil {
+		s.kubeClient, err = k8sutil.KubeClient(s.s)
+		if err != nil {
+			return false, err
+		}
+	}
+
+	err = s.kubeClient.Get(ctx, client.ObjectKey{Name: "cluster"}, &s.configV1)
```
cblecker, line 81: Why do we need to store this back in the `PodImageSpecWebhook` struct?

pkg/webhooks/podimagespec/podimagespec.go
```diff
@@ -0,0 +1,281 @@
...
+	}
+
+	for i := range pod.Spec.InitContainers {
+		container := &pod.Spec.InitContainers[i]
+		ret, err := s.authorizeOrMutateContainer(container)
+		ret.UID = request.AdmissionRequest.UID
+		if err != nil {
+			return admissionctl.Errored(http.StatusInternalServerError, err)
+		}
+	}
```
cblecker, line 125: You can only admit or reject the whole pod.. I'm not sure I understand why we are looping over multiple containers at this step? You're also overwriting the `ret` with each loop, so it will only action the last iteration of the loop.

pkg/webhooks/podimagespec/podimagespec.go
```diff
@@ -0,0 +1,293 @@
...
+
+	pod, err := s.renderPod(request)
+	if err != nil {
+		log.Error(err, "couldn't render a Pod from the incoming request")
+		return admissionctl.Errored(http.StatusBadRequest, err)
+	}
+
+	if !podContainsContainerRegexMatch(pod) {
+		return admissionctl.Allowed("Pod image spec is valid")
```
mjlshen, line 95: Any time we return a response (besides bad requests), we need to ensure the `uid` is set, matching the request. There are two patterns in this repo:
* Doing it manually: https://github.com/openshift/managed-cluster-validating-webhooks/blob/8bda2d2b7709223945b04311e604c5a2b73e4f03/pkg/webhooks/service/service.go#L80-L81
* Using a helper: https://github.com/openshift/managed-cluster-validating-webhooks/blob/8bda2d2b7709223945b04311e604c5a2b73e4f03/pkg/webhooks/imagecontentpolicies/imagecontentpolicies.go#L55 / https://github.com/openshift/managed-cluster-validating-webhooks/blob/8bda2d2b7709223945b04311e604c5a2b73e4f03/pkg/webhooks/utils/utils.go#L116-L122

pkg/webhooks/podimagespec/podimagespec.go
```diff
@@ -0,0 +1,293 @@
...
+		containerMatch, namespace, _, _ := checkContainerImageSpecByRegex(pod.Spec.InitContainers[i].Image)
+		if containerMatch && namespace == "openshift" {
+			podMatch = true
+		}
+	}
+
+	return
+}
+
+func (s *PodImageSpecWebhook) mutatePod(pod *corev1.Pod, ctx context.Context) ([]byte, error) {
```
mjlshen, line 151:
```suggestion
func (s *PodImageSpecWebhook) mutatePod(ctx context.Context, pod *corev1.Pod) ([]byte, error) {
```

nit, by convention context should be the first argument

pkg/webhooks/podimagespec/podimagespec.go
```diff
@@ -0,0 +1,293 @@
...
+	if matches == nil {
+		return false, "", "", ""
+	}
+	namespaceIndex := imageRegex.SubexpIndex("namespace")
+	imageIndex := imageRegex.SubexpIndex("image")
+	tagIndex := imageRegex.SubexpIndex("tag")
+	return true, matches[namespaceIndex], matches[imageIndex], matches[tagIndex]
+}
+
+func (s *PodImageSpecWebhook) lookupImageStreamTagSpec(imagespec string, ctx context.Context) (string, error) {
```
mjlshen, line 203:
```suggestion
func (s *PodImageSpecWebhook) lookupImageStreamTagSpec(ctx context.Context, imagespec string) (string, error) {
```

nit, by convention context should be the first argument

mjlshen (changes requested):
Looks great overall! It's important to set the UID otherwise the api server won't align the responses it gets from the webhook with the requests it has sent out.

Otherwise, just nits about putting context as the first argument

mjlshen (approved):
/lgtm

PR #308, opened 2024-06-06
Updated podimagespec to support HCP

jewzaam:
Requires getting kubeconfig to the KAS for the HCP, not the cluster where the webbhook pod is running.


mjlshen (approved):
/approve

PR #309, opened 2024-06-06
OSD-23318: Allow to modify IDMS in hosted clusters 

samanthajayasinghe:
## What 
As part of zero-egress implementation, hosted clusters IDMS needs to update with private ECR 

- This PR allows mirroring ECR's  for hosted clusters via the feature flag
- backplane service account allows to update IDMS/ICSP/ITMS

## JIRA 
https://issues.redhat.com/browse/OSD-23318

### DIFF
--- a/pkg/webhooks/imagecontentpolicies/imagecontentpolicies.go
+++ b/pkg/webhooks/imagecontentpolicies/imagecontentpolicies.go
@@ -1,6 +1,7 @@
 package imagecontentpolicies
 
 import (
+	"context"
 	"net/http"
 	"regexp"
 	"slices"
@@ -8,10 +9,13 @@ import (
 	"github.com/go-logr/logr"
 	configv1 "github.com/openshift/api/config/v1"
 	operatorv1alpha1 "github.com/openshift/api/operator/v1alpha1"
+	"github.com/openshift/managed-cluster-validating-webhooks/config"
 	"github.com/openshift/managed-cluster-validating-webhooks/pkg/webhooks/utils"
 	admissionregv1 "k8s.io/api/admissionregistration/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 )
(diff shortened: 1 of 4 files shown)


hectorakemp:
Great work Sam! this all looks good to me :). One question - why did we change the `SyncSetLabelSelector` on the webhook? Was that part of the generated `make` code? 
Happy to `lgtm` if tested :)
/lgtm

pkg/webhooks/imagecontentpolicies/imagecontentpolicies.go
```diff
@@ -208,3 +248,17 @@ func authorizeImageContentSourcePolicy(icsp operatorv1alpha1.ImageContentSourceP
 
 	return true
 }
+
+// isAllowedUserGroup checks if the user or group is allowed to perform the action
+func isAllowedUserGroup(request admission.Request) bool {
```
bmeng, line 253: Should we allow the backplane user to do this or a particular service account?

samanthajayasinghe: Updated to only allow for SA

hectorakemp:
This generally looks good to me (the CM follows the convention set out in https://github.com/openshift/managed-cluster-validating-webhooks/pull/279). 
/lgtm

mjlshen (changes requested):
/hold

I think we should adjust the approach away from a feature flag if possible. We temporarily added one as a bit of a rush to get a feature out to a subset of the fleet in the past and it would be best if we could avoid having a long-lived implementation. Some potential options:

* Do we even need to deploy this webhook to HCP clusters? I recall that editing this IDMS resource from the hosted cluster itself has no effect, so do we even care if customers can modify it? https://github.com/openshift/managed-cluster-validating-webhooks/blob/b1e1f9bc8e52a6bc55a3eed365bd2cdd8ce12d92/pkg/webhooks/imagecontentpolicies/imagecontentpolicies.go#L172-L174
* Is it feasible to allow the Kubernetes user that `hypershift-operator` is leveraging to always modify these resources instead? The customer should only ever be able to influence this behavior through OCM right and we can shift the guardrails to OCM?

pkg/webhooks/imagecontentpolicies/imagecontentpolicies.go
```diff
@@ -22,6 +27,17 @@ const (
...
+	//Allow Hypershift hosted clusters to mirror ECR
+	authorizedECRMirrors = `(^(.*).dkr.ecr.(.*).amazonaws.com)`
+
+	allowECRConfigMapName = "allow-ecr"
+)
+
+var (
+	allowedGroups = []string{
+		"system:serviceaccounts:openshift-backplane-srep",
+	}
```
mjlshen, line 40: Granting ourselves admin access isn't a part of this card - is it needed?

samanthajayasinghe:
Thanks @mjlshen for the review, As discussed, we will disable the webhooks from hosted clusters

PR #314, opened 2024-07-12
[OSD-24459] excluding user system:serviceaccount:open-cluster-management-addon-observability:endpoint-observability-operator-sa from the set of privileged serviceaccount users

aliceh:
This commit is in response to https://issues.redhat.com/browse/OHSS-35716 - a customer service account was able to modify a core managed configmap: cluster-monitoring-config. I am excluding user system:serviceaccount:open-cluster-management-addon-observability:endpoint-observability-operator-sa from the set of privileged serviceaccount users.

https://issues.redhat.com/browse/OSD-24459


moadz:
We shouldn't merge this in its current state, as it will break RHACM running on any OSD managed cluster. I've added more detail in the Jira. https://issues.redhat.com/browse/OSD-24459

PR #315, opened 2024-07-19
OSD-24266 : Disabled IDMS/ICSP validatingwebhook in HCP

samanthajayasinghe:
## Why
This PR enable ECR image registry configuration on hosted clusters 

## Issue 
https://issues.redhat.com/browse/OSD-24266

PR #316, opened 2024-07-23
Enforce package/syncset generation in CI

mjlshen:
Noticed in #315 that package generation failures don't cause CI to fail.

[OSD-24266](https://issues.redhat.com//browse/OSD-24266)


build/pr_check.sh
mjlshen, line 1: /hold

Holding PR for testing this - this should have caused PR checks to fail

PR #318, opened 2024-08-14
OSD-24275: Validate machineCIDR is contained in default ingresscontro…

nephomaniac:
This PR intends to address [OSD-24275](https://issues.redhat.com//browse/OSD-24275) adding checks to help ensure at least one of the provided 'allowedSourceRanges' subnets contains the install-config's machineCIDR. 
This includes a new role and rolebinding in the syncselectorset allowing the validation-webhook service account get() access to the install-config configmap in the kube-system namespace. 
This PR attempts to include a set of unit tests intended to check different 'allowedSourceRanges' values across different machineCIDR values across update() and create() requests. 

This ingresscontroller validation/check is intended to have the following limitations:
- Should only apply to Create, and/or Update request operations
- Should only apply to update/create requests for the "default" ingresscontroller (namespace: openshift-ingress-operator)
- This validation should not apply to requests which do not contain the "allowedSourceRanges" param/attr, or contain an empty value in the request. (Note: This does not prevent users from removing the 'allowedSourceRanges' values from an ingress controller which can result in a controller get stuck in progressing state).
- At this time this webhook is not enabled for Hypershift, and is enabled for Classic only.


pkg/webhooks/ingresscontroller/ingresscontroller.go
```diff
@@ -149,12 +176,151 @@ func (wh *IngressControllerWebhook) authorized(request admissionctl.Request) adm
 		}
 	}
 
-	ret = admissionctl.Allowed("IngressController operation is allowed")
+	/* TODO:
+	 * 1) ONLY check for privatelink clusters? How?
```
wgordon17, line 180: I don't _believe_ this only needs to apply to PrivateLink clusters, I think it could apply to _any_ cluster

pkg/webhooks/ingresscontroller/ingresscontroller.go
```diff
@@ -149,12 +176,151 @@ func (wh *IngressControllerWebhook) authorized(request admissionctl.Request) adm
...
+	 * 2) HCP vs Classic, etc.. ...any other cluster type this should apply to? Is this handled by the
+	 * HypershiftEnabled vs ClassicEnabled flags set in this module? Currently HCP disabled.
+	 * If HCP is to be enabled for allowed source ranges, should this part of a 2nd ingress validator to
+	 * allow separation of validations between cluster install types? Is there a run time method available
+	 * to this validator to determine classic vs hcp?
```
wgordon17, line 185: > HCP vs Classic, etc.. ...any other cluster type this should apply to?
Probably OpenShift Dedicated as well (unless that's covered under `Classic`)

>Is this handled by the * HypershiftEnabled vs ClassicEnabled flags set in this module? Currently HCP disabled.
I'm not sure of the impact of this 🤔

> If HCP is to be enabled for allowed source ranges, should this part of a 2nd ingress validator to allow separation of validations between cluster install types?
This validator is _only_ needed on the `default` Ingress Controller

pkg/webhooks/ingresscontroller/ingresscontroller.go
```diff
@@ -149,12 +176,151 @@ func (wh *IngressControllerWebhook) authorized(request admissionctl.Request) adm
...
+	 * 3) What other specifics should be checked here for this cidr check to be applicable?m
```
wgordon17, line 186: AFAIK, Machine CIDR is the _only_ check that would have an impact on this, so no other checks should be necessary

pkg/webhooks/ingresscontroller/ingresscontroller.go
```diff
@@ -149,12 +176,151 @@ func (wh *IngressControllerWebhook) authorized(request admissionctl.Request) adm
...
+	 */
+	// Only check for machine cidr in allowed ranges if creating or updating resource...
+	reqOp := request.AdmissionRequest.Operation
+	if reqOp == admissionv1.Create || reqOp == admissionv1.Update {
+		//TODO: Will these need to iterate over more than just the default IngressController config?
```
wgordon17, line 191: Only the `default` Ingress Controller is needed

nephomaniac:
Noting that this PR currently has all the ingressController validations under the single IngressController webhook. This  seemed to be the pattern?  However, if we want to extend this and possibly support different validations between Classic vs HCP, etc would it then make sense to separate this out? Would we potentially want additional IngressController webhooks, and/or build time flags to indicate cluster environments such as hcp vs classic, etc?

wgordon17:
@nephomaniac that's definitely a good call out, and probably an improvement worth exploring. I do believe that's probably outside the scope of OSD-24275. For this card, perhaps its "good enough" to just focus on targeting Classic for the time being

pkg/webhooks/ingresscontroller/ingresscontroller.go
michaelryanmcneill: Correct, it needs to check against any clusters.

nephomaniac:
Converted this to a draft as it seems likely this should be applied to HCP as well and changes the implementation.  In my limited HCP testing the cluster admin can edit the default ingressController's allowedSourceRanges excluding the machineCidr network causing the same/similar errors (ie blocking the console, etc).

nephomaniac:
After some discussion I believe this PR can reviewed and tested for merge. The current intention is that additional Hypershift work , if desired, is to be tracked as part of a separate effort.

pkg/webhooks/ingresscontroller/ingresscontroller.go
nephomaniac: Thanks. This comment/concern has been removed.

pkg/webhooks/ingresscontroller/ingresscontroller.go
nephomaniac: Thanks.  Updated this comment with info/dialog from OSD-24275 as well. 
The ingresscontroller webhook currently has HypershiftEnabled set to false, and ClassicEnabled set to true. I think additional dialog in OSD-24275 suggests we'll likely want to separate HCP efforts and webhook to be tracked under a different Jira.

pkg/webhooks/ingresscontroller/ingresscontroller.go
nephomaniac: Thanks. Removed/updated comment.

pkg/webhooks/ingresscontroller/ingresscontroller.go
```diff
@@ -149,12 +176,151 @@ func (wh *IngressControllerWebhook) authorized(request admissionctl.Request) adm
...
+		if ic.ObjectMeta.Name == "default" && ic.ObjectMeta.Namespace == "openshift-ingress-operator" {
+			ret := wh.checkAllowsMachineCIDR(ic.Spec.EndpointPublishingStrategy.LoadBalancer.AllowedSourceRanges)
+			ret.UID = request.AdmissionRequest.UID
+			if !ret.Allowed {
+				log.Info("Error checking minimum AllowedSourceRange", "err", ret.AdmissionResponse.String())
+			}
+			return ret
+		}
+	}
+	log.Info("############# DEBUG LOG: IngressController operation is allowed ###########")
```
tnierman, line 201:
```suggestion
```
Looks like this is debug output? Can it be removed before merging?

pkg/webhooks/ingresscontroller/ingresscontroller.go
```diff
@@ -149,12 +176,151 @@ func (wh *IngressControllerWebhook) authorized(request admissionctl.Request) adm
...
+}
+
+func (wh *IngressControllerWebhook) checkAllowsMachineCIDR(ipRanges []operatorv1.CIDR) admissionctl.Response {
+	// https://docs.openshift.com/container-platform/4.13/networking/configuring_ingress_cluster_traffic/configuring-ingress-cluster-traffic-load-balancer-allowed-source-ranges.html
+	// Note: From docs it appears a missing ASR value/attr allows all. However...
+	// once ASR values have been added to an ingresscontroller, later deleting all the ASRs can expose an issue
+	// where the IGC will remaining in progressing state indefinitely.
+	// For now return Allowed, but with a warning?
+	if ipRanges == nil || len(ipRanges) <= 0 {
+		return admissionctl.Allowed("Allowing empty 'AllowedSourceRanges'. Populate this value if operator remains in 'progressing' state")
```
tnierman, line 284:
```suggestion
		return admissionctl.Allowed("Allowing empty 'AllowedSourceRanges'")
```

IMO, we should avoid suggesting configuration changes in component logs; the product documentation should be the source-of-truth we provide to customers

pkg/webhooks/ingresscontroller/ingresscontroller.go
```diff
@@ -149,12 +176,151 @@ func (wh *IngressControllerWebhook) authorized(request admissionctl.Request) adm
...
+	}
+	log.Info("############# DEBUG LOG: IngressController operation is allowed ###########")
+	ret = admissionctl.Allowed("IngressController operation is allowed, machineCIDR n/a")
 	ret.UID = request.AdmissionRequest.UID
 
 	return ret
 }
 
+func (wh *IngressControllerWebhook) getMachineCIDR() (net.IP, *net.IPNet, error) {
+	if wh.machineCIDRIP == nil || wh.machineCIDRNet == nil {
```
tnierman, line 209: Rather than lazy-loading the CIDR on first request, would it make sense to just initialize these values in the `NewWebhook()` function? 

The benefits I see with that approach (correct me if I'm wrong on any of these points, though):
- We can omit the `kubeClient` field from the webhook, since it looks like the only time it's used is when we call `getClusterConfig`
- we'll know at hook startup if anything goes wrong w/ the retrieval, rather than at first request (which may make it slightly easier to identify issues here)
- we don't sacrifice any performance on the first request

pkg/webhooks/ingresscontroller/ingresscontroller.go
```diff
@@ -149,12 +176,151 @@ func (wh *IngressControllerWebhook) authorized(request admissionctl.Request) adm
...
+			err := fmt.Errorf("can not fetch machineCIDR from empty '%s' install config", installConfigMap)
+			log.Error(err, "getMachineCIDR failed to find CIDR value")
+			return nil, nil, err
+		}
+		if len(instConf.Networking.MachineCIDR) <= 0 {
+			err := fmt.Errorf("empty machineCIDR string value parsed from '%s' install config", installConfigMap)
+			log.Error(err, "getMachineCIDR found empty CIDR value")
+			return nil, nil, err
+		}
+		machIP, machNet, err := net.ParseCIDR(string(instConf.Networking.MachineCIDR))
```
tnierman, line 225:
```suggestion
		_, machNet, err := net.ParseCIDR(string(instConf.Networking.MachineCIDR))
```

Do we actually need both the IP and network here? Seems like we could get by on just the network, no?

pkg/webhooks/ingresscontroller/ingresscontroller.go
```diff
@@ -42,7 +54,22 @@ var (
...
+
+// installConfig struct to load the min values needed from the install-config data.
+// Alternative would be to vendor all of github.com/openshift/installer/pkg/types to import the proper struct.
+// Required values: machineCidr info, anything else we gather from this? hcp vs classic, privatelink info, etc?
+type installConfig struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata"`
+	Networking        struct {
+		MachineCIDR string `json:"machineCIDR"`
+	} `json:"networking"`
```
tnierman, line 72: Personally, I think we should just import the proper struct from `github.com/openshift/installer/pkg/types` - it doesn't seem like the package is all that large and it'll potentially simplify a lot of logic under the `getMachineCIDR()` and `getClusterConfig()` functions

pkg/webhooks/ingresscontroller/ingresscontroller.go
nephomaniac: Thanks Trevor, I think you're correct. For the validations we really only need the machNet (ie: net.ipNet) .network() value and mask size. This is mainly for debug and can be removed.
I had originally left the machIP (ie net.IP) mostly for debug purposes so it would might be easier to map a logged message or response error to the actual value found in the config map.  This might not be a practical scenario, and/or reason to keep this around though? 
ie if it's possible to build a cluster with a full IP ie: '192.168.1.1/8' (maybe accidentally?).  We'll handle and log this as '192.0.0.0/8'.  Which might make debug a little harder when searching for the stored value?

pkg/webhooks/ingresscontroller/ingresscontroller.go
nephomaniac: Thanks, will do.

pkg/webhooks/ingresscontroller/ingresscontroller.go
nephomaniac: Thanks. I've made a change making the call during the init/NewWebHook() func exit with error if we fail to load and parse the machineCIDR. This should now only fetch and parse the config once at init time. Testing this now.

pkg/webhooks/ingresscontroller/ingresscontroller.go
nephomaniac: Updated to only using ipNet in latest commit.

pkg/webhooks/ingresscontroller/ingresscontroller.go
nephomaniac: Updated in latest commit.

nephomaniac:
Ready for review. 
Updated to load + cache machineCidr, or fail fast during webhook initialization. Moved shared test args specific to webhooks into the webhooks package allowing unit tests and make target tests to run without having to access a live cluster's config.

nephomaniac:
Will need another dry-run or equiv webhook flag to allow additional make targets, build routines to iterate through hooks w/o requiring an actual cluster runtime env.

pkg/webhooks/ingresscontroller/ingresscontroller.go
tnierman: Lets start by just setting up guardrails around the `default` ingresscontroller, and if it turns out we need to check additional ICs, we can iterate.

There's little risk to this approach, IMO, since the changes here are adding new checks, so worst case scenario would be that we just don't reduce alerts as much as we expect.

nephomaniac: Thanks. Will leave the scope limited to the 'default' IC for now.

tnierman:
/approve

This looks good to me, but lets get a second opinion too

nephomaniac:
Thanks @iamkirkbater, 
The first tests will need to be things outside of the unit tests. General init/install/bootstrapping. My biggest concern is race conditions on init as I've only tested this by replacing the webhook daemon set on a running cluster.   

For pointed tests targeted specifically this validation I'd recommend the following minimum set of tests, somewhat similar to the unit tests:
Assuming a default machine cidr of 10.0.0.0/16 
- Test a request believed to be valid, including the machine cidr in one of the provided allowed subnets...
```oc -n openshift-ingress-operator patch ingresscontroller/default --type=merge --patch='{"spec":{"endpointPublishingStrategy": {"loadBalancer":{"allowedSourceRanges":["10.128.0.0/16", "10.129.0.0/16", "10.69.0.0/16", "10.0.0.0/8", "172.30.0.0/16"]}}}}' ```

- Test an invalid request where none of the subnets contain the machine cidr...
```oc -n openshift-ingress-operator patch ingresscontroller/default --type=merge --patch='{"spec":{"endpointPublishingStrategy": {"loadBalancer":{"allowedSourceRanges":["10.128.0.0/16", "10.129.0.0/16", "10.69.0.0/16",  "172.30.0.0/16"]}}}}' ```

- Test a valid request which does not contain an allowedSourceRanges field or is empty. This should be tested before and after a cluster has had the allowedSourceRanges field populated. 
_Note: https://docs.openshift.com/container-platform/4.13/networking/configuring_ingress_cluster_traffic/configuring-ingress-cluster-traffic-load-balancer-allowed-source-ranges.html
	// Note: From docs it appears a missing ASR value/attr allows all. However...
	// once ASR values have been added to an ingresscontroller, later deleting all the ASRs can expose an issue
	// where the IGC will remaining in progressing state indefinitely 
```oc -n openshift-ingress-operator patch ingresscontroller/default --type=merge --patch='{"spec":{"endpointPublishingStrategy": {"loadBalancer":{"allowedSourceRanges":[]}}}}' ```

- Repeat the above tests using a cluster with a non-default machine cidr (ie machineCidr != 10.0.0.0/16). 
- Upgrade with and without a valid ingresscontroller allowedSourceRange config (is this WH in the path?).
- Likely outside the scope of this PR, but additional testing for IGC requests that degrade  the cluster that are missed by this webhook.

bmeng:
It seems we implemented several functions within the webhook. Does it mean those functions will be called for each validation? Would it be any potential performance degradation?

nephomaniac:
> It seems we implemented several functions within the webhook. Does it mean those functions will be called for each validation? Would it be any potential performance degradation?

Thanks @bmeng, 
Currently the intention is to attempt the client/fetch work, which could have potential request delay/overhead, only once during  the init/newWebhook() portion. 
The machineCIDR value should be cache'd for use during later ingresscontroller request subnet validations.  The new subnet validation functions should be limited to 'create' & 'update' requests, and only for the 'default' ingresscontroller. Currently not enabled for hypershift. 
Worth noting that if the WH's client fails to fetch the machineCIDR during init, the WH will exit, retrying on the next init.

PR #334, opened 2024-11-06
Add osde2e tests harness

YiqinZhang:
- Add the osde2e test harness and migrated tests from osde2e. 
https://github.com/openshift/osde2e/blob/main/pkg/e2e/webhooks/webhooks.go

[SDCICD-1396](https://issues.redhat.com//browse/SDCICD-1396)

PR #335, opened 2024-11-11
misc: bump controller-runtime -> v0.15.1

jbpratt:
NewDecoder no longer returns an error and instead panics if the scheme is nil. It previously always returned a nil error

https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.14.6/pkg/webhook/admission#NewDecoder

https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.15.1/pkg/webhook/admission#Decoder.Decode

to reproduce:

```
go get sigs.k8s.io/controller-runtime@v0.15.1
```

then resolve build errors

required for #334


jbpratt:
/unhold

validated the osde2e tests are still passing with this change

PR #339, opened 2025-01-27
OSD-21711 check to allow hive admin user to trigger an SDN migration

luis-falcon:
We are adding this change to ensure that hive is the one that can trigger the SDN to OVN network migration. 

Hive does not use a service account to connect to the cluster but a privilege admin user. 

 Please Refer to [OSD-21711](https://issues.redhat.com//browse/OSD-21711)  for full explanation.

### DIFF
--- a/pkg/webhooks/sdnmigration/sdnmigration.go
+++ b/pkg/webhooks/sdnmigration/sdnmigration.go
@@ -16,9 +16,10 @@ import (
 )
 
 const (
-	WebhookName        string = "sdn-migration-validation"
-	docString          string = `Managed OpenShift customers may not modify the network config type because it can can degrade cluster operators and can interfere with OpenShift SRE monitoring.`
-	overrideAnnotation string = "unsupported-red-hat-internal-testing"
+	WebhookName               string = "sdn-migration-validation"
+	docString                 string = `Managed OpenShift customers may not modify the network config type because it can can degrade cluster operators and can interfere with OpenShift SRE monitoring.`
+	overrideAnnotation        string = "unsupported-red-hat-internal-testing"
+	privilegedHiveUserAccount string = "admin-kubeconfig-signer"
 )
 
 var (
@@ -45,6 +46,15 @@ type NetworkConfigWebhook struct {
 
 // Authorized will determine if the request is allowed
 func (w *NetworkConfigWebhook) Authorized(request admissionctl.Request) admissionctl.Response {
+	// We are doing this check to ensure that hive can trigger the
+	// migration process. Once a cluster install completes successfully,
+	// the admin password and kubeconfig will be uploaded as secrets and linked to the ClusterDeployment resource
+	// on hive under the cluster namespace. Hive uses this credentials for the user "admin-kubeconfig-signer"
+	// in order to call the api on the clusters and execute administrative tasks.
+	if request.UserInfo.Username == privilegedHiveUserAccount {
+		return utils.WebhookResponse(request, true, "Privileged user may access")
+	}
+
 	// allow if modified by an allow listed service account
 	for _, group := range request.UserInfo.Groups {
 		if privilegedServiceAccountsRe.Match([]byte(group)) {


pkg/webhooks/sdnmigration/sdnmigration.go
```diff
@@ -24,6 +24,7 @@ const (
 var (
 	log                         = logf.Log.WithName(WebhookName)
 	privilegedServiceAccountsRe = regexp.MustCompile(utils.PrivilegedServiceAccountGroups)
+	priviledgeUserAccount       = "admin-kubeconfig-signer"
```
joshbranham, line 27: Since we are specifically referring to this as the `hive` user, let's name it as such. Also type in `priviledge`, versus `priviledged`. Something like `privilegedHiveUserAccount`.

Also, this should be a constant as it's a `string`, not a `var`

pkg/webhooks/sdnmigration/sdnmigration.go
```diff
@@ -46,6 +47,10 @@ type NetworkConfigWebhook struct {
 // Authorized will determine if the request is allowed
 func (w *NetworkConfigWebhook) Authorized(request admissionctl.Request) admissionctl.Response {
 	// allow if modified by an allow listed service account
+	if priviledgeUserAccount == request.UserInfo.Username {
```
joshbranham, line 50: This reads backwards to me. We want to compare the request username to our constant, not the other way around.

pkg/webhooks/sdnmigration/sdnmigration.go
```diff
@@ -46,6 +47,10 @@ type NetworkConfigWebhook struct {
...
+		return utils.WebhookResponse(request, true, "Priviledge user may access")
```
joshbranham, line 51: Typo: `Privileged user may access` to follow the existing pattern.

joshbranham (changes requested):
Can you provide a description in the PR body of the change and it's intent?

Additionally, can you provide a comment as to why this check is added, and that Hive is using the admin credentials for the cluster to authorize itself, as to not confuse folks down the road.

pkg/webhooks/sdnmigration/sdnmigration.go
luis-falcon: Change done

pkg/webhooks/sdnmigration/sdnmigration.go
luis-falcon: Change done

pkg/webhooks/sdnmigration/sdnmigration.go
luis-falcon: change done

pkg/webhooks/sdnmigration/sdnmigration.go
```diff
@@ -45,6 +46,12 @@ type NetworkConfigWebhook struct {
 
 // Authorized will determine if the request is allowed
 func (w *NetworkConfigWebhook) Authorized(request admissionctl.Request) admissionctl.Response {
+	//We are doing this Check to endure That hive can trigger the
+	//migration process. Please refer to OSD-21711 for full explenation.
```
joshbranham, line 50: Comments should be a single space after `//`. Additionally there are typos and inconsistent casing in the comment, please review.

joshbranham (changes requested):

joshbranham (commented):
Once this deploys to stage we will want to validate this works

/lgtm
/unhold

joshbranham (commented):
/approve

pkg/webhooks/sdnmigration/sdnmigration.go
```diff
@@ -45,6 +46,12 @@ type NetworkConfigWebhook struct {
 
 // Authorized will determine if the request is allowed
 func (w *NetworkConfigWebhook) Authorized(request admissionctl.Request) admissionctl.Response {
+	// We are doing this check to ensure that hive can trigger the
+	// migration process. Please refer to OSD-21711 for full explanation.
```
fahlmant, line 50: Can you please explain some more here? This OSD card is private so not everyone can see the detai.s.

luis-falcon: Done

PR #340, opened 2025-01-27
Remove lisa as a reviewer in OWNERS

lisa:
I regret that I don't have the time to stay as attached to mcvw as I was when I ported it from Python to Go, but the sad reality is, my time is just not as available as it once was. :(


joshbranham (commented):
/lgtm

PR #346, opened 2025-02-05
Update test case to enforce security context settings

Amarthya-v:
What type of PR is this?
Test Update

What this PR does / why we need it?
This PR updates the test case for managed-cluster-validating-webhooks to enforce security context settings.

Which Jira/GitHub issue(s) this PR fixes?
Part of [OSD-28133](https://issues.redhat.com//browse/OSD-28133)

Special notes for your reviewer:
The test case now includes SecurityContext settings to enforce security policies.
This change is limited to test logic and does not affect operational code.

### DIFF
--- a/osde2e/managed_cluster_validating_webhooks_test.go
+++ b/osde2e/managed_cluster_validating_webhooks_test.go
@@ -135,11 +135,6 @@ var _ = Describe("Managed Cluster Validating Webhooks", Ordered, func() {
 		Expect(err).NotTo(HaveOccurred())
 	})
 
-	
-	func boolPtr(b bool) *bool {
-		return &b
-	}
-
 	Describe("sre-pod-validation", Ordered, func() {
 		const (
 			privilegedNamespace   = "openshift-backplane"
@@ -637,3 +632,8 @@ var _ = Describe("Managed Cluster Validating Webhooks", Ordered, func() {
 		})
 	})
 })
+
+// Helper function to create a pointer to a bool
+func boolPtr(b bool) *bool {
+	return &b
+}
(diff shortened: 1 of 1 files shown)


osde2e/managed_cluster_validating_webhooks_test.go
```diff
@@ -165,7 +194,7 @@ var _ = Describe("Managed Cluster Validating Webhooks", Ordered, func() {
 
 			err := client.Get(ctx, saName, namespaceName, sa)
 
-			if err == nil {
+			if (err == nil) {
```
joshbranham, line 197: This is not necessary, should be left as is 👍

osde2e/managed_cluster_validating_webhooks_test.go
```diff
@@ -603,3 +632,8 @@ var _ = Describe("Managed Cluster Validating Webhooks", Ordered, func() {
 		})
 	})
 })
+
+// Helper function to create a pointer to a bool
+func boolPtr(b bool) *bool {
+	return &b
+}
```
joshbranham, line 639: All usage of this can be replaced with the builtin K8s `pointer` function `pointer.BoolPtr(true)`

joshbranham (changes requested):

osde2e/managed_cluster_validating_webhooks_test.go
btoll: @Amarthya-v Please make sure that your editor is using tools like `gofmt` which will catch stuff like this.

osde2e/managed_cluster_validating_webhooks_test.go
```diff
@@ -105,6 +105,36 @@ var _ = Describe("Managed Cluster Validating Webhooks", Ordered, func() {
...
+							SeccompProfile: &v1.SeccompProfile{
+								Type: v1.SeccompProfileTypeRuntimeDefault,
+							},
+						},
+					},
+				},
+			},
+		}
+
+		// Create the pod
```
btoll, line 133: Just a nit, but I feel the code is descriptive enough here that the comment isn't needed.

joshbranham (commented):
/lgtm

PR #348, opened 2025-02-07
OSD-28159 - Enable OLM Skip Range

anispate:
Configure boilerplate to generate an OLM bundle with a skip range pointing to the latest version. This will allow OLM to upgrade past broken replaces chains in which versions get abadoned without a path forward. The new OLM bundle will contain a single version, referencing the latest build with a skipRange to the version. Part of [OSD-28159](https://issues.redhat.com//browse/OSD-28159) ticket.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -2,6 +2,7 @@ package config
 
 const (
 	// I know this isn't the operator's name but so much stuff has been coded to use this...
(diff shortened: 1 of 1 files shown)


joshbranham:
@anispate this is not an operator so this change will have no impact

PR #353, opened 2025-03-05
Added cleanup for pod and Namespace updated

Amarthya-v:
What type of PR is this?
Namespace Handling & test pod Cleanup Update

What this PR does / why we need it?
This PR improves the managed-cluster-validating-webhooks test case by:
Ensuring the test namespace exists before running tests.
Implementing automated cleanup of test resources after execution.

Key Changes:
Namespace Existence Check & Creation (BeforeAll Hook)

The test now verifies if the required namespace (testNsName) exists.
If the namespace is not found, it will be created dynamically.
This prevents test failures due to missing namespaces.

Which Jira/GitHub issue(s) this PR fixes?
Part of [OSD-28343](https://issues.redhat.com//browse/OSD-28343)

Special notes for your reviewer:
This PR only modifies test logic and does not affect operational code.
The added namespace validation ensures that tests do not fail due to missing namespaces.
The cleanup process ensures a cleaner test environment by properly removing test resources.


test/e2e/managed_cluster_validating_webhooks_test.go
```diff
@@ -320,9 +328,13 @@ var _ = Describe("Managed Cluster Validating Webhooks", Ordered, func() {
 		)
 
 		BeforeAll(func(ctx context.Context) {
-			testNamespace = &v1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: testNsName}}
-			err := client.Create(ctx, testNamespace)
-			Expect(err).ShouldNot(HaveOccurred(), "Unable to create test namespace")
+			ns := &v1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: testNsName}}
+			err := client.Get(ctx, testNsName, "", ns)
+			if errors.IsNotFound(err) {
+				By(fmt.Sprintf("Creating namespace %s", testNsName))
```
btoll, line 334: The way I'm reading this is that the `testNsName` namespace shouldn't exist before this block, so the conditional shouldn't be necessary.

It appears that the `testNsName` is being created implicitly in the `Describe("sre-pod-validation"` block, but then it looks like every `Create` call has the namespace overridden with the privileged namespace.

I did a quick test to confirm this by only running the `Describe` block mentioned above and setting a timeout between pod creation and deletion, and the `testNsName` was not created.

PR #354, opened 2025-03-07
Ensure namespace is online before creating cluster resources

btoll:
https://issues.redhat.com/browse/OSD-28343


test/e2e/managed_cluster_validating_webhooks_test.go
```diff
@@ -132,27 +134,23 @@ var _ = Describe("Managed Cluster Validating Webhooks", Ordered, func() {
...
 		var pod *v1.Pod
 
 		BeforeAll(func() {
-			name := envconf.RandomName("testpod", 12)
 			pod = &v1.Pod{
 				ObjectMeta: metav1.ObjectMeta{
-					Name:      name,
-					Namespace: testNsName,
+					Name:      envconf.RandomName("testpod", 12),
+					Namespace: privilegedNamespace,
```
btoll, line 153: I set `privilegedNamespace` as the default here because the previous value (`testNsName`) was always being overridden.  `privilegedNamespace` is used in every current test except for one, and the spec that is not using it can be addressed specific to that test.

test/e2e/managed_cluster_validating_webhooks_test.go
```diff
@@ -201,20 +199,17 @@ var _ = Describe("Managed Cluster Validating Webhooks", Ordered, func() {
 			}
 		})
 
-		withNamespace := func(pod *v1.Pod, namespace string) *v1.Pod {
-			pod.SetNamespace(namespace)
-			return pod
-		}
```
btoll, line 207: This can be removed and the code simplified by setting the `privilegedNamespace` as the default `ns` in the pod's `ObjectMeta` definition above.

test/e2e/managed_cluster_validating_webhooks_test.go
```diff
@@ -201,20 +199,17 @@ var _ = Describe("Managed Cluster Validating Webhooks", Ordered, func() {
...
-			err = userk8s.Create(ctx, withNamespace(pod, privilegedNamespace))
+			err = userk8s.Create(ctx, pod)
 			Expect(errors.IsForbidden(err)).To(BeTrue())
 
-			err = userk8s.Create(ctx, withNamespace(pod, unprivilegedNamespace))
+			pod.SetNamespace(unprivilegedNamespace)
+			err = userk8s.Create(ctx, pod)
 			Expect(errors.IsForbidden(err)).To(BeTrue())
+			pod.SetNamespace(privilegedNamespace)
 		}, SpecTimeout(createPodWaitDuration.Seconds()+deletePodWaitDuration.Seconds()))
```
btoll, line 213: This is the only test that doesn't use the `privilegedNamespace`, so I'm moved the responsibility of toggling the `ns` to it.

PR #355, opened 2025-03-07
Ensure namespace is created before creating any cluster resources

btoll:
https://issues.redhat.com/browse/OSD-28343

### DIFF
--- a/osde2e/managed_cluster_validating_webhooks_test.go
+++ b/osde2e/managed_cluster_validating_webhooks_test.go
@@ -330,6 +330,12 @@ var _ = Describe("Managed Cluster Validating Webhooks", Ordered, func() {
 			err := dedicatedAdmink8s.Delete(ctx, cm)
 			Expect(errors.IsForbidden(err)).To(BeTrue(), "Expected to be forbidden from deleting user-ca-bundle ConfigMap")
 
+			By("checking the custom namespace exists")
+			err = wait.For(conditions.New(client.Resources).ResourceMatch(testNamespace, func(object k8s.Object) bool {
+				return true
+			}))
+			Expect(err).ToNot(HaveOccurred())
+
 			cm = &v1.ConfigMap{
 				ObjectMeta: metav1.ObjectMeta{Name: "test", Namespace: testNsName},
 				Data:       map[string]string{"test": "test"},
--- a/osde2e/managed_cluster_validating_webhooks_test.go
+++ b/osde2e/managed_cluster_validating_webhooks_test.go
@@ -57,6 +57,22 @@ var _ = Describe("Managed Cluster Validating Webhooks", Ordered, func() {
 		testNsName    = "osde2e-temp-ns"
 	)
 
+	createNS := func(ns string) {
+		testNamespace = &v1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: ns}}
+		err := client.Create(context.TODO(), testNamespace)
+		By("checking the custom namespace exists")
+		err = wait.For(conditions.New(client.Resources).ResourceMatch(testNamespace, func(object k8s.Object) bool {
+			return true
+		}))
+		Expect(err).ShouldNot(HaveOccurred(), "Unable to create test namespace")
+	}
+
+	deleteNS := func(ns *v1.Namespace) {
+		err := client.Delete(context.TODO(), ns)
+		err = wait.For(conditions.New(client.Resources).ResourceDeleted(ns))
+		Expect(err).ShouldNot(HaveOccurred(), "Unable to delete test namespace")
+	}
+
 	BeforeAll(func() {
 		log.SetLogger(GinkgoLogr)
 		var err error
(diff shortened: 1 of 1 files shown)


osde2e/managed_cluster_validating_webhooks_test.go
```diff
@@ -48,15 +48,33 @@ var _ = Describe("Managed Cluster Validating Webhooks", Ordered, func() {
...
-		testNsName    = "osde2e-temp-ns"
+		namespaceName         = "openshift-validation-webhook"
+		serviceName           = "validation-webhook"
+		daemonsetName         = "validation-webhook"
+		configMapName         = "webhook-cert"
+		secretName            = "webhook-cert"
+		saName                = "webhook-sa"
+		testNsName            = "osde2e-temp-ns"
+		privilegedNamespace   = "openshift-backplane"
+		unprivilegedNamespace = "openshift-logging"
```
btoll, line 59: These last two constants were defined twice in separate blocks, so I moved them here to only define them once.

osde2e/managed_cluster_validating_webhooks_test.go
```diff
@@ -132,27 +150,25 @@ var _ = Describe("Managed Cluster Validating Webhooks", Ordered, func() {
...
 		var pod *v1.Pod
 
 		BeforeAll(func() {
-			name := envconf.RandomName("testpod", 12)
 			pod = &v1.Pod{
 				ObjectMeta: metav1.ObjectMeta{
-					Name:      name,
-					Namespace: testNsName,
+					Name:      envconf.RandomName("testpod", 12),
+					Namespace: privilegedNamespace,
```
btoll, line 171: I set privilegedNamespace as the default here because the previous value (testNsName) was always being overridden. privilegedNamespace is used in every current test except for one, and the spec that is not using it can be addressed specific to that test.

osde2e/managed_cluster_validating_webhooks_test.go
```diff
@@ -201,20 +217,17 @@ var _ = Describe("Managed Cluster Validating Webhooks", Ordered, func() {
 			}
 		})
 
-		withNamespace := func(pod *v1.Pod, namespace string) *v1.Pod {
-			pod.SetNamespace(namespace)
-			return pod
-		}
```
btoll, line 207: This can be removed and the code simplified by setting the privilegedNamespace as the default ns in the pod's ObjectMeta definition above.

osde2e/managed_cluster_validating_webhooks_test.go
```diff
@@ -201,20 +217,17 @@ var _ = Describe("Managed Cluster Validating Webhooks", Ordered, func() {
...
 
-			err = userk8s.Create(ctx, withNamespace(pod, privilegedNamespace))
+			err = userk8s.Create(ctx, pod)
 			Expect(errors.IsForbidden(err)).To(BeTrue())
 
-			err = userk8s.Create(ctx, withNamespace(pod, unprivilegedNamespace))
+			pod.SetNamespace(unprivilegedNamespace)
+			err = userk8s.Create(ctx, pod)
 			Expect(errors.IsForbidden(err)).To(BeTrue())
+			pod.SetNamespace(privilegedNamespace)
```
btoll, line 230: This is the only test that doesn't use the privilegedNamespace, so I'm moved the responsibility of toggling the ns to it.

osde2e/managed_cluster_validating_webhooks_test.go
```diff
@@ -613,10 +624,11 @@ var _ = Describe("Managed Cluster Validating Webhooks", Ordered, func() {
 			rule := newPrometheusRule(privilegedNamespace)
 			err = client.Delete(ctx, rule)
 			Expect(err == nil || errors.IsNotFound(err)).To(BeTrue(), "Failed to ensure PrometheusRule deletion")
+			createNS(testNsName)
```
btoll, line 627: This block will create its own `ns` instead of re-using a previous block's one.

osde2e/managed_cluster_validating_webhooks_test.go
```diff
@@ -320,9 +332,11 @@ var _ = Describe("Managed Cluster Validating Webhooks", Ordered, func() {
...
 
 		BeforeAll(func(ctx context.Context) {
-			testNamespace = &v1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: testNsName}}
-			err := client.Create(ctx, testNamespace)
-			Expect(err).ShouldNot(HaveOccurred(), "Unable to create test namespace")
+			createNS(testNsName)
+		})
+
+		AfterAll(func(ctx context.Context) {
+			deleteNS(testNamespace)
```
btoll, line 339: The block should cleanup any resources it creates.

osde2e/managed_cluster_validating_webhooks_test.go
```diff
@@ -132,27 +150,25 @@ var _ = Describe("Managed Cluster Validating Webhooks", Ordered, func() {
 			},
 		}
 
-		err := client.Create(context.TODO(), pod)
+		err := client.Create(ctx, pod)
+		Expect(err).NotTo(HaveOccurred())
+		err = client.Delete(ctx, pod)
```
btoll, line 155: The block wasn't cleaning up the pod.

PR #360, opened 2025-04-09
Bp update

anispate:
Update to 1.23 go version


joshbranham (commented):
/hold
What is the reason for upgrading just the e2e tests? If we are going to update we need to `make boilerplate-update` and also update the `go.mod`.

joshbranham:
Alright looks like a few things:

- This repo does _not_ use `boilerplate` conventions
- This repo is _already_ on Go 1.23 for the main module

Therefore updating the e2e manually to use a newer image is fine 👍

PR #349, opened 2025-02-18
[OSD-19341] Remove exception allowing prometheusrules to be created in 'openshift-customer-monitoring' & 'openshift-user-workload-monitoring'

reedcort:
[OSD-19341](https://issues.redhat.com/browse/OSD-19341)

Reverts an exception in the `prometheusrule webhook` allowing anyone to create `prometheusrule` objects in the `openshift-customer-monitoring` & `openshift-user-workload-monitoring` namespace. The exception was created to unblock APP-SRE migration workload monitoring off of openshift-customer-monitoring (https://issues.redhat.com/browse/APPSRE-6523 / [OSD-13680](https://issues.redhat.com/browse/OSD-13680))

PR #364, opened 2025-04-17
Add exception for openshift-customer-monitoring for the PrometheusRule webhook

reedcort:
Reverting https://github.com/openshift/managed-cluster-validating-webhooks/pull/349 change, App SRE still uses the openshift-customer-monitoring ns and removing the exception cause [errors](https://console-openshift-console.apps.rosa.appsrep09ue1.03r5.p3.openshiftapps.com/k8s/ns/app-interface-production/pods/qontract-reconcile-openshift-prometheus-rules-1-5595c6dbb79r2n6/logs) and blocking https://issues.redhat.com/browse/ASIC-629

[OHSS-43133](https://issues.redhat.com/browse/OHSS-43133)

### DIFF
--- a/pkg/webhooks/prometheusrule/prometheusrule.go
+++ b/pkg/webhooks/prometheusrule/prometheusrule.go
@@ -80,7 +80,10 @@ func (s *prometheusruleWebhook) authorized(request admissionctl.Request) admissi
 		return admissionctl.Errored(http.StatusBadRequest, err)
 	}
 
-	if hookconfig.IsPrivilegedNamespace(pr.GetNamespace()) {
+	if hookconfig.IsPrivilegedNamespace(pr.GetNamespace()) &&
+		// TODO: [OSD-13680] Remove this exception for openshift-customer-monitoring
+		pr.GetNamespace() != "openshift-customer-monitoring" &&
+		pr.GetNamespace() != "openshift-user-workload-monitoring" {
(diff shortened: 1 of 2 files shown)


pkg/webhooks/prometheusrule/prometheusrule.go
```diff
@@ -80,7 +80,10 @@ func (s *prometheusruleWebhook) authorized(request admissionctl.Request) admissi
 		return admissionctl.Errored(http.StatusBadRequest, err)
 	}
 
-	if hookconfig.IsPrivilegedNamespace(pr.GetNamespace()) {
+	if hookconfig.IsPrivilegedNamespace(pr.GetNamespace()) &&
+		// TODO: [OSD-13680] Remove this exception for openshift-customer-monitoring
+		pr.GetNamespace() != "openshift-customer-monitoring" &&
+		pr.GetNamespace() != "openshift-user-workload-monitoring" {
```
tnierman, line 86: The title and description state that appSRE only needs the `openshift-customer-monitoring` namespace, but it looks like we're re-applying the exception to the UWM namespace here as well. Is there a reason for that?

reedcort: I added it with the exception that AppSRE would migrate their monitoring to UWM in the future

joshbranham (commented):
Approving while internally we are prioritizing removing SRE management of `openshift-customer-monitoring` in https://issues.redhat.com/browse/OSD-29566

/approve

PR #365, opened 2025-04-21
copying e2e suite update from boilerplate

ritmun:
MCVW  is not boilerplated, so osde2e convention updates need to be done manually from 

https://github.com/openshift/boilerplate/tree/master/boilerplate/openshift/golang-osd-operator-osde2e


joshbranham (commented):
/lgtm

PR #375, opened 2025-07-03
Bp update leftover

ritmun:
- need to rmeove project.mk, it will come from bp
- one more bp update


joshbranham (commented):
/lgtm
/approve

PR #379, opened 2025-08-08
[WIP] [SREP-1045] Clusterrole webhook testing in stage

diakovnec:
This PR is for testing a new clusterrole webhook. We have a way to test it for Rosa Classic, but it is complicated for HCP because of image embedding and ACM policies that deploy the package to the HCP namespace. 

One of the possible ways is to merge PR into the stage and test it there. 

The functionality of the webhook itself has been verified for Rosa Classic and theoretically should work for HCP as long as it's enabled with `func (s *ClusterRoleWebHook) HypershiftEnabled() bool { return true }`

Also, the purpose of this PR is to check the e-2-e pipeline process. 

Test for Classic Rosa

```
 oc delete clusterrole system:node
Error from server (Forbidden): admission webhook "clusterroles-validation.managed.openshift.io" denied the request: Deleting ClusterRole system:node is not allowed
```
```
I0808 01:04:11.343741       1 clusterrole.go:122] "msg"="Found clusterrole: system:node" "logger"="clusterroles-validation"
I0808 01:04:11.343760       1 clusterrole.go:127] "msg"="Deleting operation detected on ClusterRole: system:node" "logger"="clusterroles-validation"

```


pkg/webhooks/clusterrole/clusterrole.go
```diff
@@ -0,0 +1,245 @@
...
+
+// isProtectedClusterRole returns true if the ClusterRole is in the protected list or matches protected patterns
+func isProtectedClusterRole(clusterRole *rbacv1.ClusterRole) bool {
+	// Check if it's in the explicit protected list (includes specific system roles)
+	if slices.Contains(protectedClusterRoles, clusterRole.Name) {
+		return true
+	}
+
+	// Check if it matches backplane pattern
+	if strings.HasPrefix(clusterRole.Name, "backplane-") {
```
bmeng, line 176: should better define this as `var` at top

pkg/webhooks/clusterrole/clusterrole.go
```diff
@@ -0,0 +1,245 @@
...
+				APIGroups:   []string{"rbac.authorization.k8s.io"},
+				APIVersions: []string{"v1"},
+				Resources:   []string{"clusterroles"},
+				Scope:       &scope,
+			},
+		},
+	}
+
+	// Protected ClusterRoles that should not be deleted
+	protectedClusterRoles = []string{
```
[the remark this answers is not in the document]
diakovnec: The full list will contain any backplane related clusterroles + most openshift critical. Its not fiesible to include all `system:*` roles as they can change over time, and we still want customers being able to create and remove their own clusterroles. The idea is being able to get access to the cluster if some clusterroles have been deleted.

pkg/webhooks/clusterrole/clusterrole.go
diakovnec: move to the top as a constant

```
const (
	WebhookName     string = "clusterroles-validation"
	backplanePrefix string = "backplane-"
```


```
// Check if it matches backplane pattern
	if strings.HasPrefix(clusterRole.Name, backplanePrefix) {
		return true
	}
```

PR #386, opened 2025-08-12
SREP-769 - Only allow 'openshift.io/cluster-monitoring: true' label to be removed from namespaces

tnierman:
https://issues.redhat.com/browse/SREP-769

---

Updates the namespace webhook to only allow customers to remove the `openshift.io/cluster-monitoring: "true"` label from their namespaces, and outright prevents them from adding or otherwise modifying it. Per [this KCS article](https://access.redhat.com/solutions/6992399), customers are not allowed to set this label on their own namespaces; doing so may cause unexpected alerts for Primary.


smarthall:
Other than that one question I have, this looks good to me.

PR #497, opened 2026-04-28
allow the delete any service account in default ns

bmeng:
We allow the serviceaccount `default` `builder` and `deployer` to be deleted in all the namespaces.

But in default namespace, there are only those 3 accounts exist from the beginning. That means every other SA in default ns should be owned by user or other operator. It does not make sense to prevent them to be deleted by user.

### DIFF
--- a/pkg/webhooks/serviceaccount/serviceaccount.go
+++ b/pkg/webhooks/serviceaccount/serviceaccount.go
@@ -51,6 +51,7 @@ var (
 		"deployer",
 	}
 	exceptionNamespaces = []string{
+		"default",
 		"openshift-logging",
 		"openshift-user-workload-monitoring",
 		"openshift-operators",


pkg/webhooks/serviceaccount/serviceaccount.go
```diff
@@ -51,6 +51,7 @@ var (
 		"deployer",
 	}
 	exceptionNamespaces = []string{
+		"default",
```
xiaoyu74, line 54: Just to understand your motivation for creating this PR: was this found from your `CI/e2e` work, or from an actual cluster case where a user/operator-created SA in the `default` namespace could not be deleted due to this webhook?

I also want to clarify the main assumption here: apart from the standard `builder`, `default`, and `deployer` SAs, any edge case that managed platform-owned SA ever be created in the `default` namespace?  If not, adding `default` to `exceptionNamespaces` looks reasonable to me.

bmeng: It comes from the CI testing, one of the OLM testing failed due to cannot clean up the serivce account in default ns.

After review the webhook, I do not see a reason to protect the service account under default is needed as none of them managed by us.

PR #612, opened 2026-08-12
Fix : [ROSAENG-61334] fail-open bypass via dispatcher mutex and unbounded body read

samanthajayasinghe:
## Summary

Fixes [ROSAENG-61334](https://issues.redhat.com/browse/ROSAENG-61334) — global dispatcher mutex + unbounded body read enables fail-open bypass of all validating webhooks.

- **Remove global mutex**: `Dispatcher.HandleRequest` held a process-wide `sync.Mutex` that serialized all admission requests. Since the `hooks` map is read-only after initialization and `hook()` returns fresh instances, no synchronization is needed. The serial bottleneck allowed a single slow/malicious request to block all others past the 2s webhook timeout, triggering `failurePolicy: Ignore` bypass.
- **Add request body size limit**: Wrap `r.Body` with `http.MaxBytesReader` (1 MiB) to reject oversized payloads that could hold connections open via unbounded `io.ReadAll`.
- **Add HTTP server timeouts**: Set `ReadHeaderTimeout` (5s), `ReadTimeout` (10s), and `WriteTimeout` (10s) on `http.Server` to prevent slowloris-style connection holding.

## Changed files

- `pkg/dispatcher/dispatcher.go` — remove mutex, add `MaxBytesReader`
- `cmd/main.go` — add server timeouts
- `pkg/dispatcher/dispatcher_test.go` — new tests for oversized body rejection, valid requests, unknown URIs, and concurrent request handling

## Test plan

- [x] `make build` passes
- [x] `make test` passes (all existing + 4 new dispatcher tests)
- [x] `make vet` passes
- [ ] Verify in staging that admission requests are processed concurrently
- [ ] Verify oversized request bodies (>1 MiB) are rejected with HTTP 400
- [ ] Verify slow clients are disconnected after server timeout

## References

- JIRA: [ROSAENG-61334](https://issues.redhat.com/browse/ROSAENG-61334)
- CWEs: CWE-770 (Allocation of Resources Without Limits), CWE-667 (Improper Locking), CWE-400 (Uncontrolled Resource Consumption)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[ROSAENG-61334]: https://redhat.atlassian.net/browse/ROSAENG-61334?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ROSAENG-61334]: https://redhat.atlassian.net/browse/ROSAENG-61334?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Added protection against request bodies larger than 3 MiB.
  * Added HTTP server timeouts to improve handling of slow or incomplete requests.

* **Bug Fixes**
  * Improved request processing reliability during concurrent traffic.
  * Requests to unknown endpoints now receive consistent responses.

* **Tests**
  * Expanded coverage for valid requests, oversized payloads, unknown endpoints, and concurrent processing.

### DIFF
--- a/pkg/dispatcher/dispatcher.go
+++ b/pkg/dispatcher/dispatcher.go
@@ -15,7 +15,7 @@ import (
 
 var log = logf.Log.WithName("dispatcher")
 
-const maxBodyBytes = 1 << 20 // 1 MiB
+const maxBodyBytes = 3 * 1024 * 1024 // 3 MiB, matches kube-apiserver MaxRequestBodyBytes default
 
 // Dispatcher struct
 type Dispatcher struct {
--- a/pkg/dispatcher/dispatcher.go
+++ b/pkg/dispatcher/dispatcher.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
-	"sync"
 
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	admissionctl "sigs.k8s.io/controller-runtime/pkg/webhook/admission"
@@ -16,10 +15,11 @@ import (
 
 var log = logf.Log.WithName("dispatcher")
 
+const maxBodyBytes = 1 << 20 // 1 MiB
+
 // Dispatcher struct
 type Dispatcher struct {
 	hooks *map[string]webhooks.WebhookFactory // uri -> hookfactory
-	mu    sync.Mutex
 }
 
 // NewDispatcher new dispatcher
@@ -41,8 +41,7 @@ func NewDispatcher(hooks webhooks.RegisteredWebhooks) *Dispatcher {
 // request, or some internal problem) it is appropriate to use the HTTP status
 // code to communicate.
 func (d *Dispatcher) HandleRequest(w http.ResponseWriter, r *http.Request) {
-	d.mu.Lock()
-	defer d.mu.Unlock()
+	r.Body = http.MaxBytesReader(w, r.Body, maxBodyBytes)
 	log.V(1).Info("Handling request", "request", r.RequestURI)
 	url, err := url.Parse(r.RequestURI)
 	if err != nil {
--- a/pkg/dispatcher/dispatcher_test.go
+++ b/pkg/dispatcher/dispatcher_test.go
@@ -2,12 +2,14 @@ package dispatcher
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"net/http"
 	"net/http/httptest"
 	"strings"
 	"sync"
 	"testing"
+	"time"
 
 	admissionv1 "k8s.io/api/admission/v1"
 	admissionregv1 "k8s.io/api/admissionregistration/v1"
@@ -39,6 +41,18 @@ func (f *fakeWebhook) SyncSetLabelSelector() metav1.LabelSelector { return metav
 func (f *fakeWebhook) ClassicEnabled() bool                       { return true }
 func (f *fakeWebhook) HypershiftEnabled() bool                    { return false }
 
+type blockingWebhook struct {
+	fakeWebhook
+	entered chan struct{}
+	release chan struct{}
+}
+
+func (b *blockingWebhook) Authorized(_ admissionctl.Request) admissionctl.Response {
+	b.entered <- struct{}{}
+	<-b.release
+	return admissionctl.Allowed("ok")
+}
+
 func newTestDispatcher() *Dispatcher {
 	hooks := webhooks.RegisteredWebhooks{
 		"test-validation": func() webhooks.Webhook { return &fakeWebhook{} },
@@ -46,7 +60,8 @@ func newTestDispatcher() *Dispatcher {
 	return NewDispatcher(hooks)
 }
 
-func validAdmissionReviewBody() []byte {
+func validAdmissionReviewBody(t testing.TB) []byte {
+	t.Helper()
 	ar := admissionv1.AdmissionReview{
 		Request: &admissionv1.AdmissionRequest{
 			UID: types.UID("test-uid"),
@@ -63,7 +78,10 @@ func validAdmissionReviewBody() []byte {
 			Operation: admissionv1.Create,
 		},
 	}
-	b, _ := json.Marshal(ar)
+	b, err := json.Marshal(ar)
+	if err != nil {
+		t.Fatalf("failed to marshal AdmissionReview: %v", err)
+	}
 	return b
 }
 
@@ -75,7 +93,7 @@ func TestHandleRequest_OversizedBody(t *testing.T) {
 		oversized[i] = 'A'
 	}
 
-	req := httptest.NewRequest("POST", "/test-hook", bytes.NewReader(oversized))
+	req := httptest.NewRequestWithContext(context.Background(), "POST", "/test-hook", bytes.NewReader(oversized))
 	req.Header.Set("Content-Type", "application/json")
 	w := httptest.NewRecorder()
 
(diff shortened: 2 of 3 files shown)


pkg/dispatcher/dispatcher.go
```diff
@@ -16,10 +15,11 @@ import (
 
 var log = logf.Log.WithName("dispatcher")
 
+const maxBodyBytes = 1 << 20 // 1 MiB
```
xiaoyu74: Could you help me understand how the 1 MiB limit was chosen here? If a AdmissionReview exceeds this local limit, we would return HTTP 400, which could then be treated as a webhook failure and skipped?

xiaoyu74:
Thanks Sam, one other question on the scope of the fix, can see the jira remediation [here](https://redhat.atlassian.net/browse/ROSAENG-61334#:~:text=Restrict%20ingress%20with%20a%20NetworkPolicy%20that%20only%20allows%20kube%2Dapiserver/host%2Dnetwork%20on%20port%205000%2C%20and%20consider%20failurePolicy%3A%20Fail%20for%20the%20high%2Dvalue%20guards%20(regular%2Duser%2C%20namespace%2C%20scc).) also mentions restricting ingress to the webhook service with a NetworkPolicy. 

I don't see a network-level restriction included in this PR, is that intentionally out of scope and planned as a follow-up?

pkg/dispatcher/dispatcher_test.go
```diff
@@ -0,0 +1,153 @@
...
+			},
+			Resource: metav1.GroupVersionResource{
+				Group:    "",
+				Version:  "v1",
+				Resource: "namespaces",
+			},
+			Operation: admissionv1.Create,
+		},
+	}
+	b, _ := json.Marshal(ar)
```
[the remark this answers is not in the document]
samanthajayasinghe: The 1 MiB limit is a pragmatic choice — the kube-apiserver itself limits request bodies to 3 MiB by default (`--max-request-body-bytes`, [ref](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/)), and the AdmissionReview payload sent to webhooks is a subset of that. Typical AdmissionReview objects are well under 1 MiB. The limit protects against an attacker sending arbitrarily large payloads to hold connections.

On the fail-open concern: when the body exceeds the limit, we return HTTP 400. The kube-apiserver treats a non-2xx response as a definitive webhook decision, not a connectivity failure — so `failurePolicy: Ignore` does not apply and the request won't be silently admitted.

pkg/dispatcher/dispatcher.go
samanthajayasinghe: The 1 MiB limit is a pragmatic choice — the kube-apiserver defaults MaxRequestBodyBytes to 3 MiB ([source](https://github.com/kubernetes/kubernetes/blob/89dfbebe2e823bd0b13e4b65b8460a058c7258a3/staging/src/k8s.io/apiserver/pkg/server/config.go#L433-L442)), which itself is 2x the 1.5 MiB etcd recommended client request size. Since the AdmissionReview payload is a subset of that, a 1 MiB cap is generous enough for any legitimate admission request while still preventing an attacker from sending arbitrarily large payloads to hold connections.

On the fail-open concern: when the body exceeds the limit, we return HTTP 400. The kube-apiserver treats a non-2xx response as a definitive webhook decision, not a connectivity failure — so failurePolicy: Ignore does not apply and the request won't be silently admitted.

samanthajayasinghe:
> Thanks Sam, one other question on the scope of the fix, can see the jira remediation [here](https://redhat.atlassian.net/browse/ROSAENG-61334#:~:text=Restrict%20ingress%20with%20a%20NetworkPolicy%20that%20only%20allows%20kube%2Dapiserver/host%2Dnetwork%20on%20port%205000%2C%20and%20consider%20failurePolicy%3A%20Fail%20for%20the%20high%2Dvalue%20guards%20(regular%2Duser%2C%20namespace%2C%20scc).) also mentions restricting ingress to the webhook service with a NetworkPolicy.
> 
> I don't see a network-level restriction included in this PR, is that intentionally out of scope and planned as a follow-up?

Good call — this repo does have a networkpolicies-validation webhook ([pkg/webhooks/networkpolicies/](https://github.com/openshift/managed-cluster-validating-webhooks/blob/master/pkg/webhooks/networkpolicies/networkpolicies.go) ) that validates tenant changes to NetworkPolicy resources, but there's no NetworkPolicy resource deployed to restrict ingress to the webhook service itself. That would be net-new work in the SelectorSyncSet/Package Operator manifests

pkg/dispatcher/dispatcher.go
xiaoyu74: > when the body exceeds the limit, we return HTTP 400. The kube-apiserver treats a non-2xx response as a definitive webhook decision, not a connectivity failure — so failurePolicy: Ignore does not apply and the request won't be silently admitted.

I’m fine with 1 MiB if we’ve validated that it’s large enough for the legitimate `AdmissionReview` payloads MCVW can receive. My main question was just to understand whether the 1 MiB value is based on that validation, rather than being only a pragmatic threshold.

otherwise I was wondering whether aligning closer to the API server’s 3 MiB limit would give us a bit more safety margin

xiaoyu74:
> > Thanks Sam, one other question on the scope of the fix, can see the jira remediation [here](https://redhat.atlassian.net/browse/ROSAENG-61334#:~:text=Restrict%20ingress%20with%20a%20NetworkPolicy%20that%20only%20allows%20kube%2Dapiserver/host%2Dnetwork%20on%20port%205000%2C%20and%20consider%20failurePolicy%3A%20Fail%20for%20the%20high%2Dvalue%20guards%20(regular%2Duser%2C%20namespace%2C%20scc).) also mentions restricting ingress to the webhook service with a NetworkPolicy.
> > I don't see a network-level restriction included in this PR, is that intentionally out of scope and planned as a follow-up?
> 
> Good call — this repo does have a networkpolicies-validation webhook ([pkg/webhooks/networkpolicies/](https://github.com/openshift/managed-cluster-validating-webhooks/blob/master/pkg/webhooks/networkpolicies/networkpolicies.go) ) that validates tenant changes to NetworkPolicy resources, but there's no NetworkPolicy resource deployed to restrict ingress to the webhook service itself. That would be net-new work in the SelectorSyncSet/Package Operator manifests

If we consider the current changes sufficient to remediate and close this finding mentioned in the jira, and the NetworkPolicy as optional defense-in-depth rather than a required follow-up, I’m fine with that.

pkg/dispatcher/dispatcher.go
samanthajayasinghe: I haven't validated 1 MiB against actual payload sizes, and for UPDATE operations the AdmissionReview includes both `object` and `oldObject`, so a large resource could exceed 1 MiB. Bumped the limit to 3 MiB to align with the kube-apiserver's MaxRequestBodyBytes default ([source](https://github.com/kubernetes/kubernetes/blob/89dfbebe2e823bd0b13e4b65b8460a058c7258a3/staging/src/k8s.io/apiserver/pkg/server/config.go#L433-L442)). This way we're guaranteed never to reject a legitimate request the apiserver would send, while still capping against unbounded reads.

